Issue #36, opened 2019-06-12
Using Session Windows (KSQL)


tlberglund:

rmoff:
if it's not happening after six years, it ain't happening…

PR #91, opened 2019-07-19
ksqlDB Session windows (#36)

rmoff:
#36


_includes/tutorials/session-windows/ksql/code/tutorial-steps/dev/check-rowtime.sql
```diff
@@ -1 +1,2 @@
+SET CLI COLUMN-WIDTH 20;
```
colinhicks: btw, in 390ae243fa88ea0cbaf4ed3eea09f94975420fd9, we added support for defining column width for the set of commands in `docker_ksql_cli_session` blocks.
e.g. https://github.com/confluentinc/kafka-tutorials/blob/cdd3747b244aa6537eb6d20b054591edfc7789d7/_data/harnesses/aggregating-count/ksql.yml#L38. This executes the `SET CLI` command behind the scenes.

This approach works, too, however.

rmoff:
@colinhicks @bbejeck I think this PR is good to go. The test is succeeding (https://confluentinc.semaphoreci.com/jobs/5ddcd8f3-9cb3-439e-bb6a-921e9d8a0fca) but the build elsewhere seems to just stop (timeout?) - but not sure why.

_data/tutorials.yaml
```diff
@@ -120,6 +120,14 @@ hopping-windows:
   status:
     ksql: enabled
 
+session-windows:
+  title: "Create session windows"
+  slug: "/create-session-windows"
+  problem: "you have time-series events in a Kafka topic, and you want to group them into variable-size, non-overlapping time intervals based on a configurable inactivity period."
+  introduction: "Given a topic of click events on a website there are various ways that we can process it. As well as simply counting the number clicks in a regular time frame (hopping or tumbling windows) we can also perform sessionization on the data. Here the length of the time window is based on the concept of a session, which is defined based on a period of inactivity. A given user might visit a website multiple times a day, but in distinct visits. We can analyse how many clicks as well as the duration of each of these sessions."
```
colinhicks:
```suggestion
  introduction: "Given a topic of click events on a website, there are various ways that we can process it. As well as simply counting the number of clicks in a regular time frame (using hopping or tumbling windows), we can also perform sessionization on the data. Here the length of the time window is based on the concept of a session, which is defined based on a period of inactivity. A given user might visit a website multiple times a day, but in distinct visits. Using session windows, we can analyze the number of clicks and duration per visit."
```

_includes/tutorials/session-windows/ksql/markup/dev/create-inputs.adoc
```diff
@@ -0,0 +1,4 @@
+The first thing we'll do is create and populate the stream of click data. An important characteristic of this data is the _timestamp_ because this is what drives the session window. KSQL can use either the Kafka message timestamp, or a field from the message value as the timestamp. In this example we'll use the latter—the _event time_ as stored in the `timestamp` field of the message value.
```
colinhicks:
```suggestion
The first thing we'll do is create and populate the stream of click data. An important characteristic of this data is the _timestamp_ because this is what drives the session window. ksqlDB can use either the Kafka message timestamp, or a field from the message value as the timestamp. In this example we'll use the latter—the _event time_ as stored in the `timestamp` field of the message value.
```

_includes/tutorials/session-windows/ksql/markup/dev/set-properties_offset.adoc
```diff
@@ -0,0 +1,5 @@
+Now that you have stream with some events in it, let's start to use them. The first thing that we'll do is ensure that the timestamp that KSQL will use for our time-based aggregation is the one that we expect. KSQL's timestamp is exposed as a column called `ROWTIME` and we can compare this to our declared `timestamp` column; they should match.
```
colinhicks:
```suggestion
Now that you have stream with some events in it, let's start to use them. The first thing that we'll do is ensure that the timestamp that ksqlDB will use for our time-based aggregation is the one that we expect. ksqlDB's timestamp is exposed as a column called `ROWTIME`, and we can compare this to our declared `timestamp` column; they should match. 
```

_includes/tutorials/session-windows/ksql/markup/dev/set-properties_cache.adoc
```diff
@@ -0,0 +1,7 @@
+Now let's see how many clicks were made in each user session (based on IP address), and also calculate the duration of the session. To do that, we issue the following transient query to aggregate the clicks grouped by the IP address. We declare that we want to use session windowing, with a gap of five minutes or more denoting the end of the previous session. It also captures the window start and end times, and uses these to calculate the session duration. 
+
+For the purposes of this example only, we are also going to configure KSQL to buffer the aggregates as it builds them. This makes the query feel like it responds more slowly, but means that you get just one row per window from which is simpler to understand the concept:
```
colinhicks:
```suggestion
For the purposes of this example only, we are also going to configure ksqlDB to buffer the aggregates as it builds them. This makes the query feel like it responds more slowly, but it means that you get just one row per window. This makes it simpler to understand the concept: 
```

_includes/tutorials/session-windows/ksql/markup/dev/print-topic.adoc
```diff
@@ -0,0 +1,11 @@
+This creates and populate a table that we could query with a `SELECT`, but also a Kafka topic, which we can examine from within KSQL using `PRINT`.
+
++++++
+<pre class="snippet"><code class="shell">{% include_raw tutorials/session-windows/ksql/code/tutorial-steps/dev/print-topic.sql %}</code></pre>
++++++
+
+Notice that the key for each message contains some strange characters that aren't quite printable. KSQL has combined the grouping key (IP address) with its window boundaries using a format that's not quite printable in this format. It should look something like this:
```
colinhicks: I didn't see the not-quite-printable effect. Was this an issue in earlier versions, perhaps?
```suggestion
Notice the key for each message. ksqlDB has combined the grouping key (IP address) with its window boundaries. It should look something like this:
```

_includes/tutorials/session-windows/ksql/markup/dev/start-cli.adoc
```diff
@@ -0,0 +1,5 @@
+To begin developing interactively, open up the KSQL CLI:
```
colinhicks:
```suggestion
To begin developing interactively, open up the ksqlDB CLI:
```

_includes/tutorials/session-windows/ksql/markup/dev/transient-query-no-buffer.adoc
```diff
@@ -0,0 +1,9 @@
+When KSQL builds aggregates it emits the values as new messages are processed, meaning that you may see intermediate results on the screen too. When `ksql.streams.cache.max.bytes.buffering` was set above this suppressed these; if you change the value to zero and re-run the above query you'll see the intermediate emissions too:
```
colinhicks:
```suggestion
When ksqlDB builds aggregates, it emits the values as new messages are processed, meaning that you may see intermediate results on the screen too. When `ksql.streams.cache.max.bytes.buffering` was set above, this suppressed these. If you change the value to zero and re-run the above query you'll see the intermediate emissions too: 
```

_includes/tutorials/session-windows/ksql/markup/dev/set-properties_cache.adoc
```diff
@@ -0,0 +1,7 @@
+Now let's see how many clicks were made in each user session (based on IP address), and also calculate the duration of the session. To do that, we issue the following transient query to aggregate the clicks grouped by the IP address. We declare that we want to use session windowing, with a gap of five minutes or more denoting the end of the previous session. It also captures the window start and end times, and uses these to calculate the session duration.
```
colinhicks:
```suggestion
Now let's see how many clicks were made in each user session (based on IP address), and also calculate the duration of the session. To do that, we issue the following transient query to aggregate the clicks grouped by the IP address. We declare that we want to use session windowing, with a gap of five minutes or more denoting the end of the previous session. The query also captures the window start and end times, and it uses these to calculate the session duration. 
```

_includes/tutorials/session-windows/ksql/markup/try_it.html
```diff
@@ -0,0 +1,23 @@
+This creates and populate a table that we could query with a `SELECT`, but also a Kafka topic, which we can examine from within KSQL using `PRINT`.
```
colinhicks:
```suggestion
This creates and populates a table that we could query with a `SELECT`. It also creates a Kafka topic, which we can examine from within ksqlDB using `PRINT`.
```

tutorials/session-windows/ksql.html
```diff
@@ -0,0 +1,10 @@
+---
+layout: tutorial
+permalink: /create-session-windows/ksql
+stack: ksql
+static_data: session-windows
+
+try_it: tutorials/session-windows/ksql/markup/try_it.html
+test_it: tutorials/session-windows/ksql/markup/test_it.html
+take_it_to_prod: tutorials/session-windows/ksql/markup/take_it_to_prod.html
```
colinhicks: Curious if these page variables are used?

rmoff: I think this is a hangover from last year when i first wrote the tutorial - I think the framework has changed a bit since then. I'll remove them.

PR #102, opened 2019-08-01
Join Two Streams (KSQL) Recipe

riferrei:
The goal of this recipe is to join two streams of events to come up with a new stream with additional insight. This is implemented using KSQL, and the scenario is the join between orders and shipments.


MichaelDrogalis:
@riferrei I think you might need to rebase against master. This patch has 1200+ files changes.

riferrei:
This PR became unmanaged due to the amount of change to make for rebasing. Closing this one and filing another.

PR #108, opened 2019-08-02
Add KSQL Aggregation / MIN MAX recipe

rmoff:


_includes/tutorials/aggregating-minmax/ksql/code/src/statements.sql
```diff
@@ -0,0 +1,11 @@
+CREATE STREAM MOVIE_SALES (title VARCHAR, release_year INT, total_sales INT)
+             WITH (KAFKA_TOPIC='movie-ticket-sales',
+                   PARTITIONS=1,
+                   VALUE_FORMAT='avro');
+
+CREATE TABLE MOVIE_FIGURES_BY_YEAR AS
+    SELECT RELEASE_YEAR,
+        MIN(TOTAL_SALES) AS MIN__TOTAL_SALES,
```
MichaelDrogalis: Why the double underscores?

_data/tutorials.yaml
```diff
@@ -73,3 +73,12 @@ tumbling-windows:
   status:
     ksql: enabled
     kstreams: enabled
+
+aggregating-minmax:
+  title: "Calculate the number of events over a period of time"
+  slug: "/tutorials/create-stateful-aggregation-minmax"
+  problem: "you have data in a Kafka topic and want to get the minimum and maximum value of a field within the events."
+  introduction: "Suppose you have a topic with events that represent ticket sales of movies. In this tutorial, we'll write a program that calculates the maximum and minimum revenue of movies by year."
```
MichaelDrogalis: The example for this one is a bit awkward because its a min/max over an aggregation. While the events themselves are just values, it does make you step back and wonder if you're missing something.

Can we switch this up to min/max duration of movies or something? Anything that doesn't imply a pre-aggregated value.

tutorials/aggregating-minmax/kstreams.html
```diff
@@ -0,0 +1,6 @@
+---
+layout: tutorial
+permalink: /create-stateful-aggregation-minmax/kstreams
+stack: kstreams
+static_data: aggregating-minmax
```
MichaelDrogalis: Needs `help_wanted: true`

_data/tutorials.yaml
tlberglund: I added a bit of narrative to explain the pre-aggregation. I think it works, but lemme know what you think. Will push in a few...

_includes/tutorials/aggregating-minmax/ksql/code/src/statements.sql
tlberglund: Robin is an old Oracle DBA. I've learned not to ask. (Seriously, since he's on PTO and we're talking about this, this is probably idiomatic among some set of database professionals. I don't mind the look, since it sets off the name of the aggregation from the column name.)

MichaelDrogalis: Ahh, that idiom actually makes sense now that you point it out. I just didn't see the pattern at first :)

PR #115, opened 2019-08-08
Aggregations: sum (KStreams)

ybyzek:
Issue #17 

Aggregations: sum (KStreams)


MichaelDrogalis:
Hey @ybyzek -- great stuff. I went through and made some changes based on convention of the other tutorials. I also adding copy editing that I did on Robin's recipe for sum'ing with KSQL. Just a few outstanding items:

1. The tutorial could use a little more discussion of what the steps are that we're taking the user through. I think @tlberglund did a really good job with this in the [stream/table join tutorial](https://kafka-tutorials.confluent.io/join-a-stream-to-a-table/kstreams.html).

2. Step (4) on the development section refers back to the example with books, but this tutorial is about movie revenue.

3. We should mention EOS somewhere in here, if not actually flip it on. It's pretty key to the whole piece.

4. We should call out what the cache buffer config does. It's an important knob to tune :)

ybyzek:
@MichaelDrogalis 

> 1. The tutorial could use a little more discussion of what the steps are that we're taking the user through. I think @tlberglund did a really good job with this in the stream/table join tutorial.
> 2. Step (4) on the development section refers back to the example with books, but this tutorial is about movie revenue.

Both of these points addressed in https://github.com/confluentinc/kafka-tutorials/pull/115/commits/6cf8777d4ab7cf160404a136cb97c055c7dc870d, please review and let me know if it needs more tweaking.

> 3. We should mention EOS somewhere in here, if not actually flip it on. It's pretty key to the whole piece.

Idempotent producer or transactions?

> 4. We should call out what the cache buffer config does. It's an important knob to tune :)

Forthcoming...

MichaelDrogalis:
This is a really great tutorial @ybyzek! Thanks for contributing!

Since a bunch of new tutorials are going in that need some information about EOS, I'll just merge this with the target branch for those and we can figure out some text for it later.

Thanks again!

PR #118, opened 2019-08-13
New: Join Two Streams (KSQL)

riferrei:
**Problem**: You have two Kafka topics with a never-ending sequence of events coming in. You want to join the stream of events from both topics to come up with a new stream with additional insight.

**Tutorial**: Suppose you have two topics containing data about orders and shipments. In this recipe, we'll write a program that joins these two streams to figure out things like which orders have been shipped, how long did it take for them to ship, as well as from which warehouse.


_includes/tutorials/joining-stream-stream/ksql/markup/dev/transient-join.adoc
```diff
@@ -0,0 +1,11 @@
+Let's cross both streams to come up with a new stream that contains more insightful information such as which orders have been shipped, which warehouse shipped them, and how long did it take to ship each order. The following query does a inner join between the orders stream and the shipments stream, and specifies a window duration of seven days to represent the SLA of one week to ship orders. This will block and continue to return results until it's limit is reached or you tell it to stop.
```
MichaelDrogalis: - Why did we use an inner join? Roughly what does it do?
- The 7 day window isn't really an SLA -- we should be more specific and talk about how it controls for the buffering space to allow joins.

_includes/tutorials/joining-stream-stream/ksql/code/tutorial-steps/dev/continuous-join.sql
```diff
@@ -0,0 +1,11 @@
+CREATE STREAM ORDERS_SHIPMENTS AS
```
MichaelDrogalis: It might be a little more accurate to call this `shipped_orders` :)

_includes/tutorials/joining-stream-stream/ksql/markup/dev/transient-join.adoc
MichaelDrogalis: I think something that's really crucial to mention somewhere is *why* a stream/stream join? Why not a stream/table join here? This is something people really get tripped up on.

_includes/tutorials/joining-stream-stream/ksql/code/src/statements.sql
```diff
@@ -0,0 +1,25 @@
+CREATE STREAM ORDERS (ORDER_ID INT, ORDER_TS VARCHAR, TOTAL_AMOUNT DOUBLE, CUSTOMER_NAME VARCHAR)
+    WITH (KAFKA_TOPIC='_orders',
+          VALUE_FORMAT='JSON',
```
MichaelDrogalis: We should stick with Avro for these examples since we're nudging people to get the power out of schemas.

_includes/tutorials/joining-stream-stream/ksql/code/src/statements.sql
```diff
@@ -0,0 +1,25 @@
...
+          TIMESTAMP='ORDER_TS',
+          TIMESTAMP_FORMAT='yyyy-MM-dd''T''HH:mm:ssX',
+          PARTITIONS=4);
+
+CREATE STREAM SHIPMENTS (SHIPMENT_ID VARCHAR, SHIP_TS VARCHAR, ORDER_ID INT, WAREHOUSE VARCHAR)
+    WITH (KAFKA_TOPIC='_shipments',
+          VALUE_FORMAT='JSON',
```
MichaelDrogalis: Same here.

_includes/tutorials/joining-stream-stream/ksql/code/src/statements.sql
```diff
@@ -0,0 +1,25 @@
...
+          TIMESTAMP='SHIP_TS',
+          TIMESTAMP_FORMAT='yyyy-MM-dd''T''HH:mm:ssX',
+          PARTITIONS=4);
```
MichaelDrogalis: We should call out somewhere that these `4` partitions have significance, and that the streams need to be co-partitioned. We should say simply exactly what that means and what happens if you don't do it. Maybe call out that this happens automatically if it's an operation that happens as a result of other streaming operations.

_includes/tutorials/joining-stream-stream/ksql/markup/dev/create-orders-stream.adoc
```diff
@@ -0,0 +1,5 @@
+First, you'll need to create a Kafka stream and its underlying topic to represent the orders:
```
MichaelDrogalis: We should call out that we're using an attribute of the event to control for time, and note that it'll be significant later.

_includes/tutorials/joining-stream-stream/ksql/markup/dev/transient-join.adoc
MichaelDrogalis: One other thing to note is that the window is using event time. This bit can be a black box when people look at s/s joins.

_includes/tutorials/joining-stream-stream/ksql/code/tutorial-steps/dev/print-output-topic.sql
```diff
@@ -0,0 +1,4 @@
+SELECT
```
MichaelDrogalis: Can we switch this up to a `PRINT` statement? The result looks a bit more clunky, but it does help people realize that the resulting stream is just a topic that can be used like any other.

riferrei:
Alright... all changed have been made @MichaelDrogalis. Ball back to you =)

MichaelDrogalis:
Excellent tutorial. Merging!

PR #119, opened 2019-08-14
Updated the Confluent Platform version to 5.3.0

riferrei:
Virtually all tutorials have this link to `_includes/shared-content/docker-install.txt` that specifies how the user will get Confluent Platform. Currently. the version used is **v5.2.1** which is outdated.

image

This PR changes it to **v5.3.0** which is in line to what we're currently using in our docker-compose.yml file.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -29,7 +29,7 @@ exclude:
   - CONTRIBUTING.md
 
 ak_javadoc_version: 23
-cp_version: v5.2.1
+cp_version: v5.3.0
 google_tag_manager_id: GTM-NBVX4Z
 
 plugins:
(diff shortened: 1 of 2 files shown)


gAmUssA:
@riferrei shall we do dependencies upgrade across all the tutorials?

riferrei:
> @riferrei shall we do dependencies upgrade across all the tutorials?

CP docker dependency wise, the only places I've found that needed update where the `docker-install.txt` file (which every tutorial points to) and in the transforming-kafka tutorial. I've changed both places so we should be good now.

Course, I did this search on the master branch. It might have other branches yet to be merged that would require the change, but that requires coordination with the tutorial owners.

gAmUssA:
@riferrei i was talking about client libraries in `build.gradle`

riferrei:
Ah, OK. that is a good idea... but I think it would make more sense to create another PR for this.

gAmUssA:
@riferrei sounds good! Lgtm!

gAmUssA:
@riferrei also for some ksql tutorials we have separate compose files.
Did you check those?

riferrei:
> @riferrei also for some ksql tutorials we have separate compose files.
> Did you check those?

I did. But again... I did it on the master branch. Everything already merged appears to be in-sync with 5.3.0.

PR #128, opened 2019-08-21
Fix AsciiDoc link

Slayv:
Add a missed quote


MichaelDrogalis:
Thanks so much @Slayv! :)

Issue #123, opened 2019-08-17
Create a KStream: JDBC source connector ->  Key:null and Value:SpecificAvro


ybyzek:
Create a KStream

Data source: JDBC source connector and Kafka Streams application with Key:null and Value:`SpecificAvro`

Comparable existing example: https://github.com/confluentinc/examples/tree/5.3.0-post/connect-streams-pipeline#example-3a-jdbc-source-connector-with-specificavro---keystringnull-and-valuespecificavro

Why is this a useful tutorial?  

* Showcases Connect + Streams integration
* JDBC connector, by default, doesn't add a key, but most Streams operations require it
* Use DSL to add the key

ybyzek:
Published https://kafka-tutorials.confluent.io/connect-jdbc-avro/kstreams.html

PR #131, opened 2019-08-28
DEVX-1039 for GH #123: Create a KStream: JDBC source connector -> Key:null and Value:SpecificAvro

ybyzek:
This PR is a working example including Connect, starting with https://github.com/confluentinc/kafka-tutorials/issues/123

Aside from general review, there are at least two known issues that need to be resolved:

1. Problem statement doesn't render.  Probably a simple typo somewhere, still haven't found it though --> RESOLVED (thanks @MichaelDrogalis )
2. Manual execution passes (key is added to the output topic), but the unit test fails (key is NOT added to the output topic).  --> RESOLVED (thanks @mikebin )

Also need to think about desirable structure given that this is the first example with Connect, and there will be more.


MichaelDrogalis:
Hey, thanks! I'll take a look at this later in the week :)

colinhicks:
@ybyzek I think this is in good shape, now. Would you mind giving my additions a second look? Essentially I added Connect to the compose config and I added assertions on the consumer logging in the Makefile, which is now invoked by CI.

A missing piece is a link on the index page to this tutorial. An earlier commit had proposed the "Aggregate data" card. I think a good option may be a new "Connect data" card. What do you think?

ybyzek:
@colinhicks changes look good.  I also made a commit with minor text edits.

> A missing piece is a link on the index page to this tutorial. An earlier commit had proposed the "Aggregate data" card. I think a good option may be a new "Connect data" card. What do you think?

Agree on  "Connect data" but I defer to you.

Issue #99, opened 2019-07-30
Kafka Streams suppress example


gAmUssA:
Provide a recipe that uses suppress on window operation.
Use case - real-time alerting. 
emit ONLY ONE alert within given window.

colinhicks:
implemented in ce05e8f

PR #144, opened 2019-09-23
Window final result #99

DivLoic:
Add the Kafka Streams tutorial: Window Final Result  (#99 )
See: [Streams DSL Doc](https://docs.confluent.io/current/streams/developer-guide/dsl-api.html#window-final-results)


DivLoic:
Please, before review this tutorial proposal considers the following remarks:

- Special Gradle tasks have been created as Java Execution  
  - In real-life applications, topic creation and schema publication are often part of an automated process rather the streaming app itself 
  - But it's a prerequisite, so the `run` task depends on `topicCreation` and `schemaPublication`
  - To explore the result a `consumeResult` tasks as been added as a black box   

- Kafka Streams as a library let you configure your app as you want. To underline it, more advance tutorials could suggest different config files examples 
  - The config file format is HCON 
  - The slf4j implementation is Logback

- To have a more interactive experience and start to see things like (grace period, late data...) the tutorial suggest to send the input to a producer through a pipe
  - It defines the produce alias `alias produce=kafka-avro-console-producer ....`
  - It creates and tails a file `touch out.txt && nohup tail -f out.txt | produce &`
  - Then write in it `echo '{"id":"000","datetime":"'$(date +%FT%T.%z)'"...' >> out.txt`

DivLoic:
tl;dr [asciicast](https://asciinema.org/a/270040)

colinhicks:
Very exciting, @DivLoic! This is shaping up to be a fantastic contribution :)

Should we plan on reviewing it once you transition the PR out of `Draft` status?

DivLoic:
Hi @colinhicks, I'm glad you appreciate this. 

The PR status is `draft` because I'd like to create the `test` and `prod` content also. You may review it already!! It will provide me with the feedback I need to complete the tutorial.

index.html
```diff
@@ -57,6 +57,7 @@ <h2 class="subtitle">Apache Kafka is a powerful, scalable, fault-tolerant publis
             <li><a href="create-stateful-aggregation-count/ksql.html">Count a stream of events</a></li>
             <li><a href="create-stateful-aggregation-sum/ksql.html">Sum a stream of events</a></li>
             <li><a href="create-stateful-aggregation-minmax/ksql.html">Find the min/max in a stream of events</a></li>
+            <li><a href="window-final-result/kstreams.html">Emit a final result from a time window</a></li>
```
DivLoic: Fixing conflict is hard 🤕! _Note_: this should be under the "**Collect data over time**" section

DivLoic: [Fix n°2](https://github.com/DivLoic/kafka-tutorials/projects/1#card-27059231)

_includes/tutorials/window-final-result/kstreams/code/src/main/java/io/confluent/developer/WindowFinalResult.java
```diff
@@ -0,0 +1,105 @@
...
+        final Config config = ConfigFactory.load();
+
+        Properties properties = new Properties();
+
+        properties.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, config.getString("bootstrap.servers"));
+        properties.put(StreamsConfig.APPLICATION_ID_CONFIG, config.getString("application.id"));
+        properties.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, StringSerde.class);
+        properties.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, StringSerde.class);
+        properties.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "latest");
+        properties.put(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, 100);
```
DivLoic: I was wrong here. I think what I was looking for actually is: 
```java
properties.put(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG, 0)
```

DivLoic: [Fix n°3](https://github.com/DivLoic/kafka-tutorials/projects/1#card-27305815)

DivLoic:
Hi 🙋‍♂️, I've just completed this tutorial. Sorry for the delay.

_includes/tutorials/window-final-result/kstreams/code/src/test/java/io/confluent/developer/WindowFinalResultTest.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    private List<ProducerRecord<Windowed<String>, Long>> readAtLeastNOutputs(int size) {
+        List<ProducerRecord<Windowed<String>, Long>> result = new ArrayList<>();
+
+        Assert.assertThat(() -> {
+            ProducerRecord<Windowed<String>, Long> record = readNext();
+            if(null != record) result.add(record);
+            return result.size();
+        }, eventuallyEval(is(size), this.testGracePeriodSize));
```
DivLoic: I'm having second thoughts about this. Since the only way to move the streaming time and get my aggregation result is to send a new event, I will get my result immediately anyway. It works fine as long as the last piped record has a timestamp greater than `window size + grace period`. There is no need to wait here 🤔.

colinhicks: I think it makes sense to leave it this way. Some tutorial users may experiment with test data that does not meet the window size + grace period criteria, causing the test to unexpectedly fail, if this logic were simplified.

DivLoic: I think, in this case, users will also have to send a final event (the same way I do at line 115 and 165) to get back their test data. Without that, tests will fail no matter what. What do you think?

DivLoic:
Hi @colinhicks, I'm really glad you like it! Now my concerns are about checking if I'm not breaking any build. How to run the test executed by the CI? I'm not really sure of what to do with `semaphore.yml`. Finally, I suspect my `action` keys in `_data/harnesses/*/*.yml` to all be setup properly.

colinhicks:
@DivLoic, you are on the right track with getting things set up for CI. This is a relatively complicated part of the tutorial process that we want to improve, so I created a branch to demonstrate the necessary changes:
https://github.com/DivLoic/kafka-tutorials/compare/ldivad-window-final-result...colinhicks:ldivad-window-final-result-makefile

Generally, I execute the harness runner for each tutorial locally in order to verify the setup. You can use this command:
```sh
make -C _includes/tutorials/window-final-result/kstreams/code tutorial
```

I have also white-listed your GitHub user name for CI builds, so further commits to this PR should indicate the result.

One remaining task is to create an assertion in the Makefile that serves as a pass/fail indication. Other tutorials use one or more `diff`s against expected output, but anything that returns a non-zero exit code to indicate failure will work.

DivLoic:
@colinhicks, thank you for the fork. This is clear to me now. Also, the following was required

Turns out my harness-runner wasn't working and I had to use the project one 
```bash
export PATH=$(pwd)/harness_runner:$PATH
```

If I'm not wrong the test output can be collected from `../harnesses/***/kstreams.yml` by
```yaml
- action: execute_async
    file: # script consumer
    stdout: tutorial-steps/dev/outputs/actual-count.txt
    render:
      skip: # true
```

colinhicks:
@DivLoic - great!

It's interesting and unexpected that you needed to add the `harness_runner` dir to your `PATH`, manually. I am curious if you had previously [installed it with pip](https://github.com/confluentinc/kafka-tutorials#5-install-the-pip-package).

Yes, adding the `stdout` key to an `action: execute_async` block will allow you to capture output, just as you noted.

DivLoic:
@colinhicks yes I did, long time ago, that was the problem.

```bash
which harness-runner
# /usr/local/bin/harness-runner
head -1 /usr/local/bin/harness-runner
# #!/usr/local/Cellar/python/3.7.4_1/bin/python3.7
```
This shebang point to a old python version that doesn't exists anyomore (rather than using `/usr/bin/env`). I've reinstall it and it works fine now! 👌Thanks!

I can't see the build detail. Can you send the error so I can try to fix it?

colinhicks:
@DivLoic, sure thing – it looks like the last build failed based on the the `diff` command:
```
1c1
< 5
---
> 4
Makefile:6: recipe for target 'tutorial' failed
```

colinhicks:
@DivLoic, I looked into the CI failures. As of 6f62719, it appears the test is actually passing. However, we see a build failed signal because the test run fails to exit within the timeout period. 

The problem appears to be in the harness runner itself, rather than any of your code. I believe the harness runner needs to block on resource cleanup before returning a non-error result. I will verify and find a fix.

Meanwhile, I will ✅this PR! Great work! And thanks for working with the poor CI visibility. I have opened #172 for future mitigations.

DivLoic:
@colinhicks, thank you for your help and your approval!

Issue #33, opened 2019-06-12
Find distinct events in a stream (KSQL)


tlberglund:
Not sure if possible with KSQL yet

riferrei:
Working on this issue \( ﾟ◡ﾟ)/

PR #187, opened 2020-01-17
Find Distinct Events in a Stream (KSQL)

riferrei:
This tutorial shows how to display unique clicks from users in a website, making sure that duplicates are removed and hence -- only distinct events are shown. This tutorial is the KSQL version of the `finding-distinct` tutorial already implemented using Kafka Streams.

This PR addresses the following issue: https://github.com/confluentinc/kafka-tutorials/issues/33

Sincerely,

-- Ricardo


_includes/tutorials/finding-distinct/ksql/markup/dev/create-inputs.adoc
```diff
@@ -0,0 +1,5 @@
+To start off the implementation of this scenario, you need to create a stream that represents the clicks from the users. Since we will be handling time in this scenario, it is important that each click contains a timestamp indicating when that click was done. The field `TIMESTAMP` will be used for this purpose.
```
colinhicks:
```suggestion
To start off the implementation of this scenario, we will create a stream that represents the clicks from the users. Since we will be handling time, it is important that each click contains a timestamp indicating when that click was done. The field `TIMESTAMP` will be used for this purpose.
```

_includes/tutorials/finding-distinct/ksql/markup/dev/set-properties.adoc
```diff
@@ -0,0 +1,5 @@
+Now that you have stream with some events in it, let's start to leverage them. The first thing to do is set the following properties to ensure that you're reading from the beginning of the stream:
```
colinhicks:
```suggestion
Now that you have a stream with some events in it, let's start to leverage them. The first thing to do is set the following properties to ensure that you're reading from the beginning of the stream:
```

_includes/tutorials/finding-distinct/ksql/markup/dev/transient-query.adoc
```diff
@@ -0,0 +1,11 @@
+Note that we called the query that find only distinct events as `DETECTED_CLICKS` and we modeled as a table, since we are performing some aggregations in the query. Then we perform a rekeying to ensure that the `IP_ADDRESS` is used as key of the record, by transforming the table back to a stream and using the `PARTITION BY` clause. To verify if everything is working as expected, run the following query:
```
colinhicks:
```suggestion
In the first statement above, we created the query that finds only distinct events, naming it `DETECTED_CLICKS`. We modeled it as a table since the query performs aggregations.

Then, with the next two statements, we specified a re-keying to ensure that the `IP_ADDRESS` is used as key of the record, by transforming the table back to a stream and using the `PARTITION BY` clause.

To verify everything is working as expected, run the following query:
```

_includes/tutorials/finding-distinct/ksql/markup/dev/transient-window.adoc
```diff
@@ -0,0 +1,11 @@
+Now let's experiment with these events. We need to create a query capable of displaying only distinct events, which means that duplicate IP addresses should be filtered out. Events are de-duped within a 2 minute window, and only unique clicks will be shown on that window. We are going to use the `LIMIT` keyword to limit the amount of records shown in the output, but if you want to play with this query and assess if it is displaying the correct results -- go ahead and remove the limit keyword.
```
colinhicks:
```suggestion
Let's experiment with these events. We need to create a query capable of displaying only distinct events, which means that duplicate IP addresses should be filtered out. Events are de-duped within a 2 minute window, and only unique clicks will be shown on that window. We are going to use the `LIMIT` keyword to limit the amount of records shown in the output. If you want to experiment with this query and assess if it is displaying the correct results, go ahead and remove the limit keyword.
```

riferrei:
> Thanks @riferrei! I left a few suggestions.

Thanks man. Your inputs are always valuable. Just fixed!

blueedgenick:
feels like the description/problem-statement for this one needs some
caveating as it doesn't really "find distinct events in a stream" - e.g. in
the example as configured, IIUC, if the same event is produced at time 1min
59 secs and again at 2min 01 secs then they won't be detected as duplicates
at all, which is quite counter-intuitive to the use-case description ?

riferrei:
@blueedgenick,

This tutorial is the equivalent to the existing one implemented with Kafka Streams:

https://kafka-tutorials.confluent.io/finding-distinct-events/kstreams.html

It provides the exact solution only using KSQL. It uses windows to emphasize the fact that we may allow duplicated clicks after 2 minutes so if a user comes back later on and click in the same link that'd be considered a brand new click. So I am OK in changing the description/problem-statement, but we may need to do this for both tutorials.

-- @riferrei

_data/tutorials.yaml
```diff
@@ -165,7 +165,7 @@ finding-distinct:
   problem: "you have events in a Kafka topic, and you want to filter out duplicate events based on a field in the event, producing a new stream of only unique events"
```
colinhicks: @riferrei I think @blueedgenick's suggestion above is worth heeding for both KStreams and KSQL flavors. I suggested a slight rewording of the shared problem statement. To me the introduction seems okay as is. What do you all think?
```suggestion
  problem: "you have events in a Kafka topic, and you want to filter out duplicate events based on a field in the event, producing a new stream of unique events per time window"
```

riferrei: Yup, make sense!

PR #212, opened 2020-02-20
DEVX-1521 Update rekey ksql tutorial to latest ksqlDB (0.7.0) syntax and output

rspurgeon:
DRAFT PR which uses a test ksqlDB image to assert the behavior of 0.7.0 behaviors, pending the bug fix for https://github.com/confluentinc/ksql/issues/4584 which will fix the packaging concern in ksqlDB.


_includes/tutorials/rekeying/ksql/code/docker-compose.yml
```diff
@@ -48,7 +51,7 @@ services:
       SCHEMA_REGISTRY_KAFKASTORE_CONNECTION_URL: 'zookeeper:2181'
 
   ksql-server:
-    image: confluentinc/cp-ksql-server:5.3.0
+    image: colinhicks/ksqldb-cli:5.5.x-latest
```
bbejeck: is this just for testing the changes?

_includes/tutorials/rekeying/ksql/code/test/output.json
```diff
@@ -4,72 +4,72 @@
       "topic": "ratings_keyed_by_id",
       "key": 294,
       "value": {
-        "id": 294,
-        "rating": 8.2
+        "ID": 294,
```
bbejeck: Just for my own education, these changes here are for KSQL changes in output correct?

_includes/tutorials/rekeying/ksql/code/docker-compose.yml
colinhicks: correct, this will be replaced with a non-test image

_includes/tutorials/rekeying/ksql/code/test/output.json
colinhicks: It looks like expected test output was changed to match the same upper-casing as is applied when using the `PRINT` command.

colinhicks:
In 7905323 I changed the spool parsing logic to handle the new prelude lines in spooled output that were added in 0.6.0+. A problem is that this applies to all ksqlDB test harnesses, including those running 5.3.x. If we want to support both the older and newer behavior, we'll need to add a property to the respective step, or make the parsing logic more sophisticated.

With the spool change, the rekey test fails because the table output format uses different column widths in 0.7.0. At first brush, this looks like a bug in ksqlDB because relatively short table headings wrap unexpectedly. I will narrow this down further. For now, I've reverted 7905323.

colinhicks:
^ related to existing bug https://github.com/confluentinc/ksql/issues/4141#issuecomment-590431927

Unfortunately it seems we'll need to work around this by changing all expected output, then alter it again when the issue is addressed.

rspurgeon:
Superseded by #214

Issue #219, opened 2020-02-28
Connect/KStreams: rename existing tutorial to differentiate from upcoming tutorial


ybyzek:
JDBC source connector does not add a key but most KStreams operations require a key.

The key can be added in one of two places:

* In Kafka Streams: captured by KT https://github.com/confluentinc/kafka-tutorials/issues/123 (done)
* In the Connect framework with an SMT: captured by KT https://github.com/confluentinc/kafka-tutorials/issues/122 (open)

However, the naming of the first tutorial (which is already done and live) is too broad and cannot be differentiated from the new tutorial.  It doesn't specify 

(a) where the key is added (KStreams vs Connect) 
(b) the format (SpecificAvro vs JSON)

This GH issue is to rename the existing tutorial to be more specific, to disambiguate from the other tutorial (and future tutorials)

Initial proposal:

* Rename the existing one at https://github.com/confluentinc/kafka-tutorials/issues/123 to `connect-jdbc-kstreams-rekey-specificavro`, with similar title
* Name the new one at https://github.com/confluentinc/kafka-tutorials/issues/122 to `connect-jdbc-rekey-smt-kstreams-json`, with similar title

ybyzek:
From @bbejeck 

> connect-jdbc-<RE-KEY OPERATOR>-rekey-<SERIALIZED-TYPE>

Which would result in

* `connect-jdbc-kstreams-rekey-specificavro`
* `connect-jdbc-smt-rekey-kstreams-json`

ybyzek:
Resolved

ybyzek:
cc: @colinhicks 

In Kafka Streams: captured by KT #123 (done)

- connect-jdbc-kstreams-rekey-specificavro
- “Add a key to a stream from a JDBC source”

In the Connect framework with an SMT: captured by KT #122 (open)

- connect-jdbc-smt-rekey-kstreams-json
- “Key a topic from Kafka Connect with an SMT”

rmoff:
I'm about to start on #122 now. My intention is for it to be a generic tutorial on setting the key in data written by a source connector (regardless of which connector, and which value format). 
So I propose 

* shortname: `connect-source-smt-set-key`
* SEO slug: `add-key-to-topic-with-kafka-connect-smt`
* Tutorial title: `Add a key to a stream from Kafka Connect`

rmoff:
[...]

PR #221, opened 2020-02-28
Connect kstreams tutorial rename

ybyzek:
Addresses https://github.com/confluentinc/kafka-tutorials/issues/219

Note: PR is incomplete until we add a redirection because the tutorial is being renamed (cc: @colinhicks )


ybyzek:
Note the new title: It reads a bit funny bit it fits on a single line.  

```
JDBC to SpecificAvro stream adding key
```

Thoughts?

_data/tutorials.yaml
```diff
@@ -150,10 +150,10 @@ serialization:
...
     kstreams: enabled
 
-connect-jdbc-specificavro:
-  title: "Ingesting data via a JDBC source connector into Kafka Streams"
-  slug: "/connect-jdbc-specificavro"
-  problem: "you have data in a table in a database, are pulling that data via a JDBC source connector (which by default does not add a key to the Kafka message), and you want to create a Kafka Stream with an added key, using SpecificAvro."
+connect-jdbc-kstreams-rekey-specificavro:
+  title: "Ingesting data via a JDBC source connector into Kafka Streams as SpecificAvro, adding a key in Kafka Streams"
+  slug: "/connect-jdbc-kstreams-rekey-specificavro"
+  problem: "you have data in a table in a database, are pulling that data via a JDBC source connector (which by default does not add a key to the Kafka message), and you want to create a Kafka Stream using SpecificAvro, letting Kafka Streams to add the key."
```
colinhicks:
```suggestion
  problem: "you have data in a table in a database, are pulling that data via a JDBC source connector (which by default does not add a key to the Kafka message), and you want to create a Kafka Stream using SpecificAvro, letting Kafka Streams add the key."
```

index.html
```diff
@@ -94,7 +94,7 @@ <h2 class="subtitle">Apache Kafka is a powerful, scalable, fault-tolerant distri
             <div>Connect data sources & sinks.</div>
           </div>
           <ul>
-            <li><a href="connect-jdbc-avro/kstreams.html">Create a stream from a JDBC source</a></li>
+            <li><a href="connect-jdbc-kstreams-rekey-specificavro/kstreams.html">JDBC to SpecificAvro stream adding key</a></li>
```
colinhicks: I suggest leaving out the "adding key" part for now. Can address later if this becomes an important distinction from other Connect tutorials.
image

ybyzek: @colinhicks the next tutorial I'd like to write will be similar to this one, but the keying functionality will be via Connect/SMT versus Kafka Streams. If we leave off "adding key", it will be harder to differentiate the tutorials even though they are very different workflows.  WDYT?

colinhicks: @ybyzek, that makes sense. The space constraints are tough, here. Kind of like a newspaper.

Do you think it's reasonable to use just enough distinguishing detail to frame the story of each tutorial? For example, maybe the next tutorial you described can have a headline (link text) like:
`Key a stream with Kafka Connect SMT` or
`Key a stream with transformations`

ybyzek: @colinhicks  We can't just decrease the font like the classifieds in a newspaper? j/k :)

The suggestions above do leave off two details:

1. SpecificAvro: the next KT was going to be JSON, but the description can explain that .  And it may not be as high priority to mention in the titl
2. JDBC: that it's a JDBC source connector versus another source connector, but that's also not as high priority, it's a detail actually

But all in all, given the space constraints, I think you found the best solution, thank you!

colinhicks: Haha, we could always link-farm ourselves with `kafka.craigslist.org` 😛

ybyzek:
@bbejeck @colinhicks just wanted to confirm whether I may merge this PR?  Please note one Semaphore failure but it appears to be on an unrelated tutorial

colinhicks:
@ybyzek, yes, feel free to merge

Issue #217, opened 2020-02-27
Remove committed output artifacts


colinhicks:
For many, if not most, tutorials, we have committed files which were generated by a test run. These should not be source controlled. The corresponding expectation files should be, however.

A resolution should remove the exiting committed files and, for example, add a gitignore in each `_includes/...{tutorial}/` root.

PR #224, opened 2020-02-28
DEVX-1547: Updates aggregating-count for ksqlDB for 0.7 version

bbejeck:


_includes/tutorials/aggregating-count/ksql/code/tutorial-steps/dev/outputs/persistent-query/output-0.log
```diff
@@ -1,4 +1,4 @@
-Message                   
----------------------------
- Table created and running 
----------------------------
+Message
```
bbejeck: I've noticed some of these output files are checked in, but get deleted and re-created with each run.  Should we start to not include these files under source control?   Ditto for the other output log files in this PR.

colinhicks: Yeah, we opened #217 regarding this. Let's come back through later and remove these. Fine to leave for the moment.

PR #231, opened 2020-03-02
Adapts rekeying-function to ksqlDB 0.7.1

colinhicks:


_includes/tutorials/rekeying-function/ksql/code/tutorial-steps/dev/continuous-regex.sql
```diff
@@ -0,0 +1,8 @@
+CREATE STREAM customers_with_area_code AS
+    SELECT 
+      customers.rowkey as id,
+      firstname,
+      lastname,
+      phonenumber,
+      REGEXREPLACE(phonenumber, '\\(?(\\d{3}).*', '$1') as area_code
+    FROM customers;
```
colinhicks: this intermediate stream is necessary in 0.7.0+, when expression support was added to `PARTITION BY` 

cf. https://github.com/confluentinc/ksql/pull/4032

rspurgeon:
Not certain what causes this difference, but when I run the DESCRIBE FUNCTION REGEXREPLACE, I get a different jar path:

```
ksql> DESCRIBE FUNCTION REGEXREPLACE;

Name        : REGEXREPLACE
Overview    : Replace string using a regex
Type        : SCALAR
Jar         : /etc/ksql/ext/code-0.0.1.jar
Variations  :

        Variation   : REGEXREPLACE(input VARCHAR, regex VARCHAR, replacement VARCHAR)
        Returns     : VARCHAR
        Description : regexReplace string using a regex
        input       : If null, then function returns null.
        regex       : If null, then function returns null.
        replacement : If null, then function returns null.
```

colinhicks:
Yeah, that's interesting. I'm not sure how to account for the jar name difference either. I think it's innocuous though, in this case.

rspurgeon:
@colinhicks this tutorial references `compile 'io.confluent.ksql:ksql-udf:5.3.0'`

Issue #122, opened 2019-08-17
Create a KStream: JDBC source connector with SMT -> Key:Long and Value:JSON


ybyzek:
Create a KStream

Data source: JDBC source connector using SMTs that generate Key:Long and Value:JSON

Comparable existing example: https://github.com/confluentinc/examples/tree/5.3.0-post/connect-streams-pipeline#example-2-jdbc-source-connector-with-single-message-transformations---keylong-and-valuejson

PR #242, opened 2020-03-06
Correction to KT name for  (#221)

ybyzek:
This PR is a correction to https://github.com/confluentinc/kafka-tutorials/pull/221

The Tutorial is mostly fine but the name that shows up in the card is wrong.

The reason I messed this up: both KTs that I intended to write (#122 and #123) use SMTs, `setschemametadata`.  But the one I wrote so far doesn't use SMTs for rekey.  The file names are correct `connect-jdbc-kstreams-rekey-specificavro`, but when I saw the SMT, I mislabelled it as the SMT rekeyer

### DIFF
--- a/index.html
+++ b/index.html
@@ -94,7 +94,7 @@
             <div>Connect data sources & sinks.</div>
           </div>
           <ul>
-            <li><a href="connect-jdbc-kstreams-rekey-specificavro/kstreams.html">Key a stream with transformations</a></li>
+            <li><a href="connect-jdbc-kstreams-rekey-specificavro/kstreams.html">Key a stream in the app</a></li>
           </ul>
         </div>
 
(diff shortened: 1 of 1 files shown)


colinhicks:
Just to check my understanding, once #122 is implemented, are you suggesting we would have the following?
- Key a stream in the app
- Key a stream with transformations

My suggestion in https://github.com/confluentinc/kafka-tutorials/pull/221#discussion_r385832826 was a bit different.

I was thinking to use
- Key a stream with transformations or

colinhicks:
Also, the redesign's font changes mean the one-line title width argument is now moot 😄.

Still my sense is we should treat these links like headlines, providing enough detail to identify and differentiate each tutorial, in a form that corresponds to an intent a reader might have.

ybyzek:
@colinhicks 

> I was thinking to use either:
> Key a stream with transformations, or
> Key a stream with Kafka Connect SMT

"transformations" may be ambiguous -- do you mean Kafka Streams transformations?  It's too close to SMT, people may think "transformations" means SMT, not KStreams

PR #292, opened 2020-03-19
Use ValueTransformerWithKey instead of Transformer

inponomarev:
Hello, 

the current version of the tutorial uses `Transformer`. However, since `Transformer` may change the key, its usage will lead to redundant repartitioning when grouping operations are used aftewards.

I propose to rewrite this example using `ValueTransformer[WithKey]`, which will not cause the unwanted repartitioning.

By the way, it seems to me that we have a bug in `transformValues`: see the `filter((k, v) -> v != null)` after it. Without this filter, `null` values appear downstream, contrary to what is said in JavaDoc:

> If the return value of `ValueTransformerWithKey#transform(Object, Object)`  is `null`, no records are emitted.

### DIFF
--- a/_includes/tutorials/finding-distinct/kstreams/code/build.gradle
+++ b/_includes/tutorials/finding-distinct/kstreams/code/build.gradle
@@ -32,9 +32,9 @@ apply plugin: "com.github.johnrengelman.shadow"
 dependencies {
     compile "org.apache.avro:avro:1.8.2"
     implementation "org.slf4j:slf4j-simple:1.7.26"
-    implementation "org.apache.kafka:kafka-streams:2.3.0"
+    implementation "org.apache.kafka:kafka-streams:2.4.1"
     implementation "io.confluent:kafka-streams-avro-serde:5.3.1"
-    testCompile "org.apache.kafka:kafka-streams-test-utils:2.3.0"
+    testCompile "org.apache.kafka:kafka-streams-test-utils:2.4.1"
     testCompile "junit:junit:4.12"
 }
 
--- a/_includes/tutorials/finding-distinct/kstreams/code/build.gradle
+++ b/_includes/tutorials/finding-distinct/kstreams/code/build.gradle
@@ -32,9 +32,9 @@ apply plugin: "com.github.johnrengelman.shadow"
 dependencies {
     compile "org.apache.avro:avro:1.8.2"
     implementation "org.slf4j:slf4j-simple:1.7.26"
-    implementation "org.apache.kafka:kafka-streams:2.4.1"
+    implementation "org.apache.kafka:kafka-streams:2.3.0"
     implementation "io.confluent:kafka-streams-avro-serde:5.3.1"
-    testCompile "org.apache.kafka:kafka-streams-test-utils:2.4.1"
+    testCompile "org.apache.kafka:kafka-streams-test-utils:2.3.0"
     testCompile "junit:junit:4.12"
 }
 
--- a/_includes/tutorials/finding-distinct/kstreams/code/src/main/java/io/confluent/developer/FindDistinctEvents.java
+++ b/_includes/tutorials/finding-distinct/kstreams/code/src/main/java/io/confluent/developer/FindDistinctEvents.java
@@ -21,7 +21,7 @@ import io.confluent.kafka.serializers.AbstractKafkaAvroSerDeConfig;
 import io.confluent.kafka.streams.serdes.avro.SpecificAvroSerde;
 import org.apache.kafka.streams.kstream.KeyValueMapper;
 import org.apache.kafka.streams.kstream.Produced;
-import org.apache.kafka.streams.kstream.Transformer;
+import org.apache.kafka.streams.kstream.ValueTransformerWithKey;
 import org.apache.kafka.streams.processor.ProcessorContext;
 import org.apache.kafka.streams.state.StoreBuilder;
 import org.apache.kafka.streams.state.Stores;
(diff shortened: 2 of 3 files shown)


inponomarev:
Concerning the need of `filter`: it seems to be [KAFKA-9533](https://issues.apache.org/jira/browse/KAFKA-9533), which is going to be fixed soon.

bbejeck:
Hi @inponomarev, 

Thanks for the PR. 

Yes, you are correct in the usage of a `Transformer` vs. `TransformValues`, but since the tutorial is working as intended what do you think about adding your comments above to the `_includes/tutorials/finding-distinct/kstreams/markup/dev/make-topology.adoc` file?

> Concerning the need of filter: it seems to be KAFKA-9533, which is going to be fixed soon

Thanks for pointing that out.  But it seems the JavaDocs were in error, and this merged PR https://github.com/apache/kafka/pull/8298 corrects the issue.  So you'll still need to use a `filter` operator to eliminate any `null` values that result.

inponomarev:
Hi @bbejeck sorry for delay with this, I have updated the tutorial. 

Concerning the nulls in `TransformValues`, I found out that in our code we have cases when we emit `null`s  from `TransformValues` explicitly (in order to delete values from tables afterwards), so fixing the Javadoc and leaving `TransformValues` behaviour unchanged seems to be the most convenient way.

_includes/tutorials/finding-distinct/kstreams/markup/dev/make-topology.adoc
```diff
@@ -1,6 +1,9 @@
...
+Focusing on the `buildTopology` method, note how the Kafka Streams topology relies on a `https://docs.confluent.io/current/streams/javadocs/org/apache/kafka/streams/kstream/ValueTransformerWithKey.html[ValueTransformerWithKey]` and a `https://docs.confluent.io/current/streams/javadocs/org/apache/kafka/streams/state/WindowStore.html[Window Store]` to filter out the duplicate IP addresses.   Events are de-duped within a 2 minute window, and unique clicks are produced to a new topic named `distinct-clicks`.
+
+NOTE: Note that we are using `https://docs.confluent.io/current/streams/javadocs/org/apache/kafka/streams/kstream/ValueTransformerWithKey.html[ValueTransformerWithKey]` here instead of `https://docs.confluent.io/current/streams/javadocs/org/apache/kafka/streams/kstream/Transformer.html[Transformer]` since we need keys to transform data, but there is no need to re-key the stream.
```
bbejeck, line 5: Unfortunately, call-outs don't render correctly at the moment.  Can you remove the `NOTE:` text? maybe you could do something like `**_Note that we are using_**` instead.  We working on adding support for call-outs.

_includes/tutorials/finding-distinct/kstreams/code/src/main/java/io/confluent/developer/FindDistinctEvents.java
```diff
@@ -30,224 +30,227 @@
...
+        this.createTopics(envProps);
+
+        final KafkaStreams streams = new KafkaStreams(topology, streamProps);
+        final CountDownLatch latch = new CountDownLatch(1);
+
+        // Attach shutdown handler to catch Control-C.
+        Runtime.getRuntime().addShutdownHook(new Thread("streams-shutdown-hook") {
+            @Override
+            public void run() {
+                streams.close();
```
bbejeck, line 242: Suggestion - I know this is pre-existing code, but can we update `close()` to `close(Duration.ofSeconds(5))`

bbejeck:
validated by going through the tutorial steps and running the test harness locally.

_includes/tutorials/finding-distinct/kstreams/markup/dev/make-topology.adoc
inponomarev: Oh I see! I got used to Asciidoctor in my own projects, that's why I'm using its features without thinking about render support :-)

inponomarev:
@bbejeck done as you suggested, apparently I also forgot to commit `build.gradle` with version updates. Will you please re-check it with newer versions, 'it works on my machine' 😄

bbejeck:
> @bbejeck done as you suggested, apparently I also forgot to commit `build.gradle` with version updates. Will you please re-check it with newer versions, 'it works on my machine' 😄

Ack and will do. Thanks for the quick response.

_includes/tutorials/finding-distinct/kstreams/code/build.gradle
```diff
@@ -32,9 +32,9 @@ apply plugin: "com.github.johnrengelman.shadow"
 dependencies {
     compile "org.apache.avro:avro:1.8.2"
     implementation "org.slf4j:slf4j-simple:1.7.26"
-    implementation "org.apache.kafka:kafka-streams:2.3.0"
+    implementation "org.apache.kafka:kafka-streams:2.4.1"
```
bbejeck, line 35: With this change I get the following when trying to run the app at step 6

```
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/kafka/clients/admin/Admin
	at org.apache.kafka.streams.KafkaStreams.<init>(KafkaStreams.java:557)
	at io.confluent.developer.FindDistinctEvents.runRecipe(FindDistinctEvents.java:235)
	at io.confluent.developer.FindDistinctEvents.main(FindDistinctEvents.java:225)
Caused by: java.lang.ClassNotFoundException: org.apache.kafka.clients.admin.Admin
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:419)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:352)
	... 3 more
```
Maybe just revert to `2.3.0` for now?

inponomarev: Yes, no time to investigate it right now... reverted to `2.3.0`.

bbejeck:
Thanks @inponomarev for the contribution.  I've verified the tutorial steps and ran the test harness locally.  I'm going to see about running the build for the PR before merging, but it will get merged soon regardless.

Thanks again!

PR #301, opened 2020-03-24
MINOR[WIP]: Sort files before diff comparison

bbejeck:
Trying sorting of output before running `diff` command to see if this stabilizes the tests.


bbejeck:
I'll try running the test for `ksql aggregation min/max` several  times to see what the outcome is

bbejeck:
The first test run passed, kicking off again.  I'm thinking if this passes after 10 runs we can merge and use the same approach for other tests that are flaky for the same reason.

bbejeck:
Ran the semaphore build 10 times and the test passed each time.  Closing this in favor of #305

PR #305, opened 2020-03-25
MINOR:Sort output before diff in KSQL aggregation tutorials

bbejeck:


bbejeck:
The experiment on #305 worked for `aggregating-min-max` with 10 test runs, so this PR added the same changes to the other KSQL aggregation (sum, count).

bbejeck:
tests passed for the ksql version of aggregations-count, aggregations-sum, and aggregations-min-max.  I'll kick off another test run.

PR #316, opened 2020-04-07
Minor: enable ksqlDB server logging

vcrfxia:
The ksqlDB Kafka Tutorials examples currently use the `log4j-rolling.properties` file for the ksqlDB server. This properties file does not log to standard out so no ksqlDB server logs are shown when examples are run, which (1) is incongruous with the other components and (2) makes it difficult to diagnose issues when something goes wrong (e.g., the server is unable to start). 

This PR updates the docker-compose files to use the "log4j.properties" file instead, which means ksqlDB server logs will be shown as part of `docker-compose up`.

### DIFF
--- a/_includes/tutorials/aggregating-count/ksql/code/docker-compose.yml
+++ b/_includes/tutorials/aggregating-count/ksql/code/docker-compose.yml
@@ -60,7 +60,7 @@ services:
       - "8088:8088"
     environment:
       KSQL_CONFIG_DIR: "/etc/ksql"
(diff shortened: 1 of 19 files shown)


vcrfxia:
Hey @MichaelDrogalis , is the build failure on "ksqlDB Connect SMT key tests" in this PR meaningful? I see it's been failing in the master build ever since it was added about a week ago.

Should I wait to merge this PR until it's fixed and we see a clean PR build?

MichaelDrogalis:
I'm actually not sure. I think @rmoff may have added that one. You can go ahead and merge, it doesn't look like this will affect the build status.

PR #318, opened 2020-04-09
ksqlDB/Connect SMT: Avoid harness exception caused by duplicate file handlers

colinhicks:
Fixes CI test failures for this tutorial.

This PR skips writing the statements.sql file, which is also read by a separate step.
The file output was never used by the harness, so effective behavior does not change.

Also, this adds a new line to the end of the expected PRINT results, to match spooled output.

Tests pass in latest run:
https://confluentinc.semaphoreci.com/jobs/4984c4ab-acea-4757-b250-8c6216e00aac


colinhicks:
Merging this based on validation in CI.

PR #319, opened 2020-04-10
DEVX-1701: Make sure gradle procs closed

bbejeck:
During test harness execution for kstream Kafka Tutorials, the Gradle process will persist running when the test harness job completes, which seems to indirectly cause a test failure.  The test itself will pass, but since the process prevents the job from exiting, it times out, creating a failure.

### DIFF
--- a/_data/harnesses/connect-add-key-to-source/kstreams.yml
+++ b/_data/harnesses/connect-add-key-to-source/kstreams.yml
@@ -197,3 +197,8 @@ prod:
         - action: skip
           render:
             file: tutorials/connect-add-key-to-source/kstreams/markup/prod/launch-container.adoc
+
+        - action: execute
+          file: tutorial-steps/dev/clean-up.sh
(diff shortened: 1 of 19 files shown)


colinhicks:
@bbejeck what happens if I'm running a separate gradle process, unrelated to the tutorial execution? I'm imagining I'm running the gradle daemon for some other development task. That process gets killed, too right?

bbejeck:
>@bbejeck what happens if I'm running a separate gradle process, unrelated to the tutorial execution? I'm imagining I'm running the gradle daemon for some other development task. That process gets killed, too right?

You mean locally, then yes, but the test that was passing before just failed on a timeout again. 

So it looks like the fix was due to timing more than anything else.  I'm going to close this PR and remove the `kill gradle` from the stateful naming tutorial.

bbejeck:
Some tests are still timing out, so this PR is not a fix.

PR #295, opened 2020-03-20
DEVX-1628: KStreams DSL choose output topic at runtime

bbejeck:
#294

PR #328, opened 2020-04-14
MINOR: update docker-compose templates to support networks

bbejeck:
This PR updates the `docker-compose.yml` templates to support docker networking

### DIFF
--- a/templates/ksql/filtered/code/docker-compose.yml
+++ b/templates/ksql/filtered/code/docker-compose.yml
@@ -11,6 +11,8 @@ services:
     environment:
       ZOOKEEPER_CLIENT_PORT: 2181
       ZOOKEEPER_TICK_TIME: 2000
+    networks:
+      -cp
 
   broker:
     image: confluentinc/cp-enterprise-kafka:<CP-VERSION>
@@ -35,6 +37,9 @@ services:
       CONFLUENT_METRICS_REPORTER_TOPIC_REPLICAS: 1
       CONFLUENT_METRICS_ENABLE: 'true'
       CONFLUENT_SUPPORT_CUSTOMER_ID: 'anonymous'
+    networks:
+      - cp
+
 
   schema-registry:
     image: confluentinc/cp-schema-registry:<CP-VERSION>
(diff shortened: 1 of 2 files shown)


colinhicks:
This seems to be more about naming the network that docker-compose creates, and less about _supporting networks_ – they are already supported by all our current compose files, which create a network with a random name.

Is the goal to have a common Docker network name across all tutorials? Does this PR propose other advantages?

bbejeck:
@colinhicks - good point.  

I'm somewhat neutral on this PR as it's a follow-up from comments on https://github.com/confluentinc/kafka-tutorials/pull/295 regarding changes to docker-compose.

@gAmUssA any additional comments?

gAmUssA:
@colinhicks Motivation for specific network name - we can reference this name in the section when we're talking about docker image. Without consistent network naming, it would be hard to write a reproducible harness scenario.

colinhicks:
Thanks for the explanation, @gAmUssA. I'm still a bit unclear on when we would refer to a Docker network name in the tutorial text content. Is there an example where we are currently doing this – or should be ideally? I think I overlooked the specific comments in #295 where this came up.

> Without consistent network naming, it would be hard to write a reproducible harness scenario.

Is there something that is currently not working because the network name is inconsistent?

One of the nice things about docker-compose is that it abstracts some of the complexity of dealing with Docker. It seems like its default networking behavior has served the tutorials well so far. My suggestion is that we only add explicit networking configuration in cases where it makes the tutorial easier and/or less error-prone for the reader, overall.

gAmUssA:
@colinhicks check this one https://kafka-tutorials.confluent.io/aggregating-average/kstreams.html#launch-the-container

colinhicks:
Thanks, @gAmUssA. I understand now :). This change does make sense for the cases where we are building separate a Docker image for the application. The application container needs to join the network created by the docker-compose config, which is easiest to do when the network has an explicit name.

rspurgeon:
It's possible that having the same network name across all tutorials could pose some issues or confusion for users wanting to run multiple tutorials.  This may not be an issue for us as we specify explicit service names, but I just wanted to broach the idea of obtaining the network name at runtime instead of statically defining it.  One way may be using `yq` which requires a small image pull: 

```
docker run --rm -v "${PWD}":/workdir mikefarah/yq yq r docker-compose.yml 'networks.cp.name'
cp_network
```

we could repackage `yq` into a cnfl controlled image if need be.  Just an idea to float...

rspurgeon:
I keep getting reminders on this stale PR... can we close it or move forward?  @bbejeck

bbejeck:
closing for now and we can re-visit at a later time if we feel the need.

gAmUssA:
What's required to make this default?

Issue #336, opened 2020-04-30
Ksql tutorials: `TIMESTAMPTOSTRING` should always be passed a timezone


big-andy-coates:
If we want examples to work in different timezones, then any use of `TIMESTAMPTOSTRING` should include a timezone, e.g. 

```sql
SELECT o.rowkey AS order_id,
           TIMESTAMPTOSTRING(o.rowtime, 'yyyy-MM-dd HH:mm:ss') AS order_ts,
           o.total_amount,
           o.customer_name,
           s.shipment_id,
           TIMESTAMPTOSTRING(s.rowtime, 'yyyy-MM-dd HH:mm:ss') AS shipment_ts,
           s.warehouse, (s.rowtime - o.rowtime) / 1000 / 60 AS ship_time
    FROM orders o INNER JOIN shipments s
    WITHIN 7 DAYS
    ON o.rowkey = s.order_id;
```

Should become:

```
SELECT o.rowkey AS order_id,
           TIMESTAMPTOSTRING(o.rowtime, 'yyyy-MM-dd HH:mm:ss', 'UTC') AS order_ts,
           o.total_amount,
           o.customer_name,
           s.shipment_id,
           TIMESTAMPTOSTRING(s.rowtime, 'yyyy-MM-dd HH:mm:ss', 'UTC') AS shipment_ts,
           s.warehouse, (s.rowtime - o.rowtime) / 1000 / 60 AS ship_time
    FROM orders o INNER JOIN shipments s
    WITHIN 7 DAYS
    ON o.rowkey = s.order_id;
```

Otherwise the output data will not match.

PR #337, opened 2020-04-30
Updated KSQL tutorials - 0.8.x

big-andy-coates:
fixes: https://github.com/confluentinc/kafka-tutorials/issues/336

Tested KSQL joins tutorials against 0.8.1 and found some issues, which are fixed below:

- PRINT command no longer needs the topic name quoting: it's now case sensitive. (fixed across all examples)
- TIMESTAMPTOSTRING should be passed a timezone for consistent results (fixed only for join tutorials - other tutorials use this and should be updated: https://github.com/confluentinc/kafka-tutorials/issues/336)
- Added explicit ROWKEY columns to all CT/CS statements, as this is to be encouraged.
- Fixed key format output of PRINT that was outputting MIXED, which is not a valid output.
- Fixed bad path used when running ksql testing tool (fixed across repo)


colinhicks:
Much appreciated @big-andy-coates. I will review this later today.

PR #366, opened 2020-05-05
git ignore output directories & -d for ksql

big-andy-coates:
update the template and all ksql tutorials to:
 1. use `-d` switch on `docker-compose up` so that test output is more easily grok'd
 2. switch site rendering to using the _expected output_, not intermediate _actual output_ files that are generated by the test run.
 2. add `.gitignore` to ignore output directories and delete them.


templates/ksql/filtered/code/.gitignore
```diff
@@ -0,0 +1,4 @@
+extensions/
+tutorial-steps/dev/outputs/
+tutorial-steps/test/outputs/
+tutorial-steps/prod/outputs/
```
colinhicks, line 4: I think this file is extraneous. We can remove it, right?

`templates/ksql/filtering/code/.gitignore` was also created.

big-andy-coates: I was under the assumption that this 'filtered' directory is the template used to create new tutorials, while `_includes/tutorials/filtering`, (which is what I assume you're referring to, though not the different root), is a tutorial instance.

I'm going to merge this PR - if I'm wrong and this file is superfluous, then please feel free to remove it!

PR #418, opened 2020-06-04
update rekey join tutorial inline with changes to ksqlDB master

big-andy-coates:
This is a larger change than the other updates as the recent changes to key handling made the existing tutorial not really work. It was relying on the fact that a stream without a key column has `ROWKEY` set to `null` if you issue a push query.

Recent changes mean that a stream without an explicit key column have not key column, i.e. no `ROWKEY`, so I've updated the tutorial to use `DESCRIBE` to see the schema of the data and ``PRINT` to view the raw data in Kafka.

Note, needs https://github.com/confluentinc/ksql/pull/5552 to pass.


_includes/tutorials/rekeying/ksql/markup/dev/continuous-rekey.adoc
```diff
@@ -1,4 +1,4 @@
-Let's fix that. Using KSQL's appropriately named `PARTITION BY` clause we can apply a key to the messages and write it to a new stream. Here we'll use the movie identifier, `ID`. Issue the following to create a new stream that is continuously populated by its query:
```
rmoff, line 1: s/KSQL/ksqlDB

rmoff:
Thanks Andy. What version is this going to drop in?

big-andy-coates:
> Thanks Andy. What version is this going to drop in?

0.10 / 6.0

_includes/tutorials/rekeying/ksql/markup/dev/continuous-rekey.adoc
big-andy-coates: This text is deleted ;)

PR #419, opened 2020-06-04
update rekey function tutorial inline with changes to ksqlDB master

big-andy-coates:


_includes/tutorials/rekeying-function/ksql/markup/dev/continuous-rekey.adoc
```diff
@@ -1,4 +1,6 @@
-Now, using ksqlDB's appropriately-named `PARTITION BY` clause, we can apply the results of `REGEXREPLACE` as the key for each message and write the result to a new stream:
+Now, using ksqlDB's appropriately-named `PARTITION BY` clause, we can apply the results of `REGEXREPLACE` as the key for each message and write the result to a new stream.
+As we're not partitioning by an existing column, ksqlDB will add a new column to the `customers_by_area_code` stream to how the result of our function call.
```
colinhicks, line 2:
```suggestion

As we're not partitioning by an existing column, ksqlDB will add a new column to the `customers_by_area_code` stream based on the result of our function call.

```

PR #420, opened 2020-06-05
update aggregation count tutorial inline with changes to ksqlDB master

big-andy-coates:


big-andy-coates:
Ooops - wrong button press! Merged this by mistake. Let's hope the build pasess!

If you spot any issues please let me know and I'll address in a follow up PR.

PR #421, opened 2020-06-05
update aggregation min-max tutorial inline with changes to ksqlDB master

big-andy-coates:


_includes/tutorials/aggregating-count/ksql/markup/dev/transient-query.adoc
```diff
@@ -1,6 +1,6 @@
-With our test data and configuration parameters in place, let’s try a query to compute our ticket totals. A `SELECT` statement all by itself in ksqlDB is called a transient query, meaning that after we stop it, it is gone and will not keep processing the input stream. That's what we're doing in this step. The counterpart to a transient query is a _persistent_ query, which we'll create a few steps from now.
+With our test data and configuration parameters in place, let’s try a query to compute our ticket totals. A `SELECT` statement with an `EMIT CHANGES` in ksqlDB is called a transient push query, unlike a traditional db query, which returns a single result set, the query continuously streams back results and once we stop it, it is gone and will not keep processing the input stream. That's what we're doing in this step. The counterpart to a transient query is a _persistent_ query, which we'll create a few steps from now.
```
colinhicks, line 1:
```suggestion
With our test data and configuration parameters in place, let’s try a query to compute our ticket totals. A `SELECT` statement with an `EMIT CHANGES` in ksqlDB is called a transient push query. 

Unlike a traditional database query, which returns a single result set, the query continuously streams back results. Once we stop it, it is gone and will not keep processing the input stream. That's what we're doing in this step. The counterpart to a transient query is a _persistent_ query, which we'll create a few steps from now.
```

_includes/tutorials/aggregating-minmax/ksql/markup/dev/persistent-query.adoc
```diff
@@ -1,4 +1,4 @@
...
+Since the output looks right, the next step is to make the query persistent. This looks exactly like the push query, except we have added a `CREATE TABLE AS` statement to the beginning of it. This statement returns to the CLI prompt right away, having created a persistent stream processing program running in the ksqlDB engine, continuously processing input records and updating the resulting `MOVIE_FIGURES_BY_YEAR` table. Moreover, we don't see the results of the query displayed in the CLI, because they are updating the newly-created table itself. That table is available to other ksqlDB queries for further processing, and by default all its records are produced to a topic having the same name (`MOVIE_FIGURES_BY_YEAR`).
```
colinhicks, line 1:
```suggestion
Since the output looks right, the next step is to make the query persistent. This looks exactly like the push query, except we have added a `CREATE TABLE AS` statement to the beginning of it. This statement returns to the CLI prompt right away, having created a persistent stream processing program running in the ksqlDB engine, continuously processing input records and updating the resulting `MOVIE_FIGURES_BY_YEAR` table.

Moreover, we don't see the results of the query displayed in the CLI, because they are updating the newly-created table itself. That table is available to other ksqlDB queries for further processing, and by default all its records are produced to a topic having the same name (`MOVIE_FIGURES_BY_YEAR`).
```

_includes/tutorials/aggregating-minmax/ksql/markup/dev/transient-query.adoc
```diff
@@ -1,6 +1,6 @@
...
 
-If you're minimally familiar with SQL, the text of the query itself is fairly self-explanatory. We are calculating the highest and lowest grossing movie figures by year using `MIN` and `MAX` aggregations on the `TOTAL_SALES` field. This query will keep running, continuing to return results until you hit CTRL-C. Most ksqlDB queries are _continuous queries_ that run forever in this way; there is always potentially more input available in the source stream, so the query never finishes on its own.
+If you're minimally familiar with SQL, the text of the query itself is fairly self-explanatory. We are calculating the highest and lowest grossing movie figures by year using `MIN` and `MAX` aggregations on the `TOTAL_SALES` column. This query will keep running, continuing to return results until you hit CTRL-C. Most ksqlDB queries are _continuous queries_ that run forever in this way; there is always potentially more input available in the source stream, so the query never finishes on its own.
```
colinhicks, line 3:
```suggestion
If you're familiar with SQL, the text of the query itself is fairly self-explanatory. We are calculating the highest and lowest grossing movie figures by year using `MIN` and `MAX` aggregations on the `TOTAL_SALES` column. This query will keep running, continuing to return results until you hit CTRL-C. Most ksqlDB queries are _continuous queries_ that run forever in this way; there is always potentially more input available in the source stream, so the query never finishes on its own.
```

PR #423, opened 2020-06-05
update connect tutorial inline with changes to ksqlDB master

big-andy-coates:
This includes adding a new SMT to remove the `city_id` from the value. This is needed if we want to name the primary key in ksqlDB `city_id`. Without it, there would be a name clash with the PK and the `city_id` column in the value schema.

Also added some details of _what_ the SMTs are doing, as its not necessarily clear to users.


_includes/tutorials/connect-add-key-to-source/ksql/markup/dev/create-connector.adoc
```diff
@@ -7,7 +7,10 @@ Launch the ksqlDB CLI:
 
 From the ksqlDB prompt you can create the JDBC source connector. There are a couple of points to note: 
 
-1. The `transforms` stanza which is responsible for setting the key to the value of the `city_id` field
+1. The `transforms` stanza which is responsible for setting the key to the value of the `city_id` field. They run in the order define by `transforms`:
```
colinhicks, line 10:
```suggestion
1. The `transforms` stanza, which is responsible for setting the key to the value of the `city_id` field. They run in the order defined by `transforms`:
```

_includes/tutorials/connect-add-key-to-source/ksql/markup/dev/create-connector.adoc
```diff
@@ -7,7 +7,10 @@ Launch the ksqlDB CLI:
...
+    * - `copyFieldToKey` sets the key to a struct containing the `city_id` field from the value.
+    * - `extractKeyFromStruct` sets the key to just the `city_id` field of the struct set by the previous step.
+    * - `removeKeyFromValue` removes the `city_id` from the message value, as its now stored in the message key.
```
colinhicks, line 13:
```suggestion
    * - `removeKeyFromValue` removes the `city_id` from the message value, as it's now stored in the message key.
```

_includes/tutorials/connect-add-key-to-source/ksql/markup/dev/create-table.adoc
```diff
@@ -1,6 +1,6 @@
 Now that we have a topic with data in from the source system and the keys appropriately set, we can declare a ksqlDB table over the topic.
 
-We only need to specify the datatype of the table's key (`ROWKEY`) - the rest of the schema is picked up automagically from the Schema Registry since we're using Avro to serialize the value part of the payload. 
+We only need to specify the datatype of the table's key (`CITY_ID`)- the rest of the schema is picked up automagically from the Schema Registry since we're using Avro to serialize the value part of the payload.
```
colinhicks, line 3:
```suggestion
We only need to specify the datatype of the table's key (`CITY_ID`) – the rest of the schema is picked up automagically from the Schema Registry since we're using Avro to serialize the value part of the payload.
```

PR #424, opened 2020-06-05
update tumbling windows tutorial inline with changes to ksqlDB master

big-andy-coates:


_includes/tutorials/transforming/ksql/markup/dev/transient-query.adoc
```diff
@@ -1,4 +1,4 @@
-Let's break apart the `title` field and extract the year that the movie was published into its own column. Issue the following transient query. This will block and continue to return results until it's limit is reached or you tell it to stop.
+Let's break apart the `title` field and extract the year that the movie was published into its own column. Issue the following transient push query. This will block and continue to return results until it's limit is reached or you tell it to stop.
```
colinhicks, line 1:
```suggestion
Let's break apart the `title` field and extract the year that the movie was published into its own column. Issue the following transient push query. This will block and continue to return results until its limit is reached or you tell it to stop.
```

_includes/tutorials/tumbling-windows/ksql/markup/dev/transient-query.adoc
```diff
@@ -1,4 +1,4 @@
-As a bonus, we can prove to ourselves that the window boundaries are in fact 6 hour intervals. Run the following transient query, which uses the `TIMESTAMPTOSTRING` function to convert the UNIX timestamps into something that we can read:
+As a bonus, we can prove to ourselves that the window boundaries are in fact 6 hour intervals. Run the following transient push query, which uses the `TIMESTAMPTOSTRING` function to convert the UNIX timestamps into something that we can read:
```
colinhicks, line 1:
```suggestion
As a bonus, we can prove to ourselves that the window boundaries are in fact 6-hour intervals. Run the following transient push query, which uses the `TIMESTAMPTOSTRING` function to convert the UNIX timestamps into something that we can read:
```

_includes/tutorials/tumbling-windows/ksql/markup/dev/transient-window.adoc
```diff
@@ -1,4 +1,4 @@
-Let's figure out how many ratings were given to each movie in 6 hour tumbling intervals. To do that, we issue the following transient query to aggregate the ratings grouped by the movie's name. This tells ksqlDB that you only want to sum up the ratings on a per movie basis. It also captures the window start and end times. These functions describe the boundaries that represent each 6 hour interval. The following will block and continue to return results until its limit is reached or you tell it to stop.
+Let's figure out how many ratings were given to each movie in 6 hour tumbling intervals. To do that, we issue the following transient push query to aggregate the ratings grouped by the movie's name. This tells ksqlDB that you only want to sum up the ratings on a per movie basis. It also captures the window start and end times. These functions describe the boundaries that represent each 6 hour interval. The following will block and continue to return results until its limit is reached or you tell it to stop.
```
colinhicks, line 1:
```suggestion
Let's figure out how many ratings were given to each movie in tumbling, 6-hour intervals. To do that, we issue the following transient push query to aggregate the ratings, grouped by the movie's name. This tells ksqlDB that you only want to sum up the ratings on a per-movie basis. It also captures the window start and end times. These functions describe the boundaries that represent each 6-hour interval. The following will block and continue to return results until its limit is reached or you tell it to stop.
```

Issue #484, opened 2020-07-26
Add extra prod step for all tutorials


ybyzek:
Add (optional) link to Confluent Cloud

ybyzek:
<img width="456" alt="Screen Shot 2020-07-26 at 16 50 01 PM" src="https://user-images.githubusercontent.com/12448600/88489261-4d11a200-cf61-11ea-8407-ebcb6295eb70.png">
<img width="1395" alt="Screen Shot 2020-07-26 at 16 56 49 PM" src="https://user-images.githubusercontent.com/12448600/88489262-4d11a200-cf61-11ea-9c63-1931acd7ec26.png">

ybyzek:
Proposal 2: move step to be last part of `prod` (versus separate section)

ybyzek:
Resolved by https://github.com/confluentinc/kafka-tutorials/pull/485

PR #485, opened 2020-07-27
GH-484: Add extra prod step for all tutorials

ybyzek:
PR for https://github.com/confluentinc/kafka-tutorials/issues/484


_data/harnesses/aggregating-average/kstreams.yml
```diff
@@ -169,3 +169,8 @@ prod:
           render:
             skip: true
 
+    - title: Run your app to Confluent Cloud
+      content:
+        - action: skip
+          render:
+            file: tutorials/change-topic-partitions-replicas/ksql/markup/prod/try-ccloud.adoc
```
rspurgeon, line 176: Does KT support a shared "include" location or can we create one easily.  Having this same file shared across tutorials might be confusing for a future reader.  I trained myself to look through these paths to be sure they point to the files in the same tutorial.  I'm worried it could be misinterpreted as a mistake if they point "across tutorials".  

cc @bbejeck @colinhicks

ybyzek: @rspurgeon similar concerns.  

Options:

1. As current: All KT refer to the same file from one KT
2. Create a dummy/shared/generic KT: all KT refer to the same file in this shared folder
3. Re-create the same physical file in all individual KTs
4. Re-create the same physical file in all individual KTs but softlink to the same file from one KT

rspurgeon: I tagged Bill & Colin because they might know more about the capabilities of the tooling.  I would guess the system might support includes somehow, but I'm not familiar with the tech.

ybyzek: @rspurgeon unless I'm missing some subtle differentiation in your suggestion, then "support includes somehow" is by definition supported with `_includes/`, see option 2.  All we have to do is agree on the directory structure.

rspurgeon: I don't know what the requirements are for the tools though, if you just add folders does it cause any issues? Haven't tested

ybyzek: @rspurgeon I will test and update the PR if it works.

colinhicks: Each render block's file path is resolved from the `_includes` directory, so it is feasible to share common files as @ybyzek suggests in #2 above.

While the convention has been to create unique files for each tutorial, there is no technical mechanism preventing the use of common files.

ybyzek: FWIW I pushed an update that implements option 2.  If we like this option, we just need to agree on subdirectory structure, and I can modify as needed.

rspurgeon: the structure you derived LGTM

bbejeck: I also agree with option 2 from the list above.

bbejeck:
@ybyzek thanks for this PR.  Once this is merged, I'll do a follow-on PR to update the template files with this new section.

ybyzek:
cc: @gAmUssA for review

ybyzek:
@bbejeck @rspurgeon as discussed, the latest commit in this PR creates a distinct step for Confluent Cloud (pulled out of Production), for all Kafka Tutorials.  Please let me know what you think with the latest change.

rspurgeon:
@ybyzek is there any way to see the staged website from this PR?  I can see the semaphore job, but don't know how to find a staged site link if one is built

_data/harnesses/splitting/kstreams.yml
```diff
@@ -174,3 +174,9 @@ prod:
           file: tutorial-steps/dev/clean-up.sh
           render:
             skip: true            
+
```
bbejeck, line 177: This one is missing the `ccloud:` and `steps`

_data/harnesses/transforming/kafka.yml
```diff
@@ -178,3 +178,9 @@ prod:
           file: tutorial-steps/dev/clean-up.sh
           render:
             skip: true
+
```
bbejeck, line 181: same here

_data/harnesses/tumbling-windows/ksql.yml
```diff
@@ -110,3 +110,9 @@ prod:
           file: tutorial-steps/dev/clean-up.sh
           render:
             skip: true
+
```
bbejeck, line 113: ditto

_data/harnesses/tumbling-windows/kstreams.yml
```diff
@@ -183,3 +183,9 @@ prod:
           file: tutorial-steps/dev/clean-up.sh
           render:
             skip: true
+
```
bbejeck, line 186: ditto

_data/harnesses/splitting/kstreams.yml
ybyzek: @bbejeck good catch, thanks, addressed.

_data/harnesses/transforming/kafka.yml
ybyzek: @bbejeck good catch, thanks, addressed.

_data/harnesses/tumbling-windows/ksql.yml
ybyzek: @bbejeck good catch, thanks, addressed.

_data/harnesses/tumbling-windows/kstreams.yml
ybyzek: @bbejeck good catch, thanks, addressed.

Issue #487, opened 2020-07-27
add important settings to finding-distinct-events ksqldb example


blueedgenick:
The way the ksqlDB example for  this use-case is written depends for correct processing on having `cache.max.bytes.buffering` and/or `commit.interval.ms` set to 0. Would be good to mention this - perhaps in the taake-to-prod section ? - to avoid negative surprises when testing at scale.

PR #488, opened 2020-07-27
Document cache.max.bytes.buffering in finding-distinct events

colinhicks:
addresses #487 

Called attention to the usage in step 3 of `Try it`, plus `Take it to production`.

[Preview link](http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/GH-487/6b17aa6e5956c6fc4d6bbb83c7a6a0aa9425309e/finding-distinct-events/ksql.html).

### DIFF
--- a/_data/harnesses/finding-distinct/ksql.yml
+++ b/_data/harnesses/finding-distinct/ksql.yml
@@ -51,6 +51,10 @@ dev:
               render:
                 file: tutorials/finding-distinct/ksql/markup/dev/set-properties.adoc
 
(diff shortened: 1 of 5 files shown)


blueedgenick:
Thanks for the super-fast edit here Colin!  LGTM, one suggestion: given the
impact that zero-sized cache can have on throughout, would it be better to
change it only for this one query (using the `set x=y` functionality)
rather than changing it globally via the compose file?

colinhicks:
thanks for the reviews!

PR #509, opened 2020-08-13
🧪 Updating jekyll and Jekyll-asciidoc to address some security issues

gAmUssA:


gAmUssA:
@bbejeck whenever you have a chance, could you release the staging website so we can click it through to make sure that rendering is ok after the update

bbejeck:
I pulled down this PR and verified by building locally

gAmUssA:
@bbejeck can we merge then?

bbejeck:
> @bbejeck can we merge then?

Probably,  but as an extra sanity check if we can get one or both of  @rspurgeon and @ybyzek to check out this PR and run `bundler install` and run the KT site locally with no issues, that would be ideal.

PR #533, opened 2020-09-08
Update deserializer errors and merging tests outputs

spena:
Updates:
- The `DESCRIBE EXTENDED` command is including the consumer group from the queries running.
- The processing log message has a new field `serializationError` that is added in the json message.


colinhicks:
Seems like `"target":"value"` is missing from `deserializationError` in `expected-print.log`

vcrfxia:
> Seems like `"target":"value"` is missing from `deserializationError` in `expected-print.log`

The change that added the `target` field is only present on master, not 0.12. Looks like we'll need a separate PR to update master after this one.

colinhicks:
Nice catch @spena, can you rebase your branch onto `ksqldb -0.12.0`. That branch specifies the RC: https://github.com/confluentinc/kafka-tutorials/blob/ksqldb-0.12.0/_includes/tutorials/deserialization-errors/ksql/code/docker-compose.yml#L53

colinhicks:
We should get the kafka tutorials tests passing against the RC, then follow up later to adjust for changes to master since the release was cut.

spena:
All tests passed. I will merge this without approved as it only updated test expected outputs.

PR #536, opened 2020-09-15
README tweaks & nits

rmoff:
Minor tweaks to fix things that glitched as I re-familiarised myself with the doc :)


bbejeck (approved):
Thanks for the improvements @rmoff!

LGTM

Issue #538, opened 2020-09-16
Create tl;dr answer one-liner at top of each KT


ybyzek:
Suggestion: Create tl;dr answer one-liner at top of each KT

Value:

1. For users who know they want the quick answer one-liner
2. Helps users find the information in the tutorial (which can be verbose)
3. SEO question/answer format
4. Potential: Minimum required KT is the one-liner answer, the rest of tutorial can be developed asynchronously

ybyzek:
<img width="1487" alt="Screen Shot 2020-09-16 at 11 23 07 AM" src="https://user-images.githubusercontent.com/12448600/93358036-00cf3980-f80f-11ea-87e1-1182a3c53bee.png">

rmoff:
+1 I think this is a good idea. 

The verbosity of the tutorials is a strength but for sure there are several of them (probably not all) where a tl;dr would really help. 

I wonder if it wouldn't make sense on all of them though. A tl;dr should be genuinely short (one sentence, maybe two), and not just a bleeding paragraph lifted from the article content.

ybyzek:
<img width="524" alt="Screen Shot 2020-09-16 at 21 05 25 PM" src="https://user-images.githubusercontent.com/12448600/93407520-68fa3b80-f860-11ea-851c-2e86a3b75cf3.png">

<img width="1470" alt="Screen Shot 2020-09-16 at 21 05 20 PM" src="https://user-images.githubusercontent.com/12448600/93407529-6d265900-f860-11ea-9c9b-390c558e555b.png">

ybyzek:
<img width="1506" alt="Screen Shot 2020-09-16 at 21 21 52 PM" src="https://user-images.githubusercontent.com/12448600/93408511-c42d2d80-f862-11ea-8a0e-222376f5174d.png">

<img width="696" alt="Screen Shot 2020-09-16 at 21 21 40 PM" src="https://user-images.githubusercontent.com/12448600/93408514-c68f8780-f862-11ea-9217-e852d4f80d18.png">

ybyzek:
@rmoff @gAmUssA @bbejeck @awalther28 @rspurgeon @colinhicks @tlberglund -- we discussed today moving the Answer to be right below the Question.  I investigated this and discovered two concerns with this approach:

[...]

PR #540, opened 2020-09-16
GH-538: Add support for "Short answer" 

ybyzek:
Addresses https://github.com/confluentinc/kafka-tutorials/issues/538

Screen Shot 2020-10-12 at 08 26 55 AM

### DIFF
--- a/_data/harnesses/aggregating-count/ksql.yml
+++ b/_data/harnesses/aggregating-count/ksql.yml
@@ -1,3 +1,11 @@
+answer:
+  steps:
+    - title: Persistent query using COUNT
+      content:
+        - action: skip
+          render:
+            file: tutorials/aggregating-count/ksql/markup/answer/count-persistent.adoc
+
 dev:
   steps:
     - title: Initialize the project
--- a/_data/harnesses/aggregating-count/kstreams.yml
+++ b/_data/harnesses/aggregating-count/kstreams.yml
@@ -1,3 +1,11 @@
+answer:
+  steps:
+    - title: count() method
+      content:
+        - action: skip
+          render:
+            file: tutorials/aggregating-count/kstreams/markup/answer/count-method.adoc
+
 dev:
   steps:
     - title: Initialize the project
--- a/_data/harnesses/console-consumer-produer-basic/kafka.yml
+++ b/_data/harnesses/console-consumer-produer-basic/kafka.yml
@@ -1,11 +1,11 @@
 answer:
   steps:
-    - title: Produce
+    - title: Console producer
       content:
         - action: skip
           render:
             file: tutorials/console-consumer-produer-basic/kafka/markup/answer/produce.adoc
-    - title: Consume
+    - title: Console consumer
       content:
         - action: skip
           render:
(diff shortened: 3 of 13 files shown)


ybyzek:
Latest staging: http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/GH-538/426e1352ce653272b2d5e1420aeafad25bf41bea/create-stateful-aggregation-count/ksql.html

ybyzek:
Latest commit uses the background suggested by Ashlee.

Screen Shot 2020-10-09 at 11 58 48 AM

Staging: http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/GH-538/97851317c3283dd60c6c7c895a17f6f0c2ff3aae/create-stateful-aggregation-count/ksql.html

Based on meeting today, all were in favor.  @bbejeck would you want to provide final approval?

gAmUssA:
can we just keep `TL;DR` without `(Short answer)`? it's just awkward. I understand SEO side of things. but still

rmoff:
+1 to what @gAmUssA said. `tl;dr (short answer)` looks clumsy. By adopting `tl;dr` we're going after the developer colloquial terminology, and then ruin it with the parentheses. 
If we're worried about the comprehension of the term we could always link to an explanation of it with a little asterisk or something like that, or a hover-over tool tip, or something.

ybyzek:
- Latest commit changes `tl;dr` to `TL;DR`.
- Reasons to keep `(Short answer)` are comprehension of term & SEO (as discussed on the call).  I lean towards keeping it because not everyone knows what it means, and it shouldn't take a link out to explain it.

gAmUssA:
can we rather wrap a link to https://en.wikipedia.org/wiki/TL%3BDR for `TLDR` and remove short answer?

gAmUssA:
we always can opt in for «Executive Summary»

bbejeck:
>Reasons to keep (Short answer) are comprehension of term & SEO (as discussed on the call). I lean towards keeping it because not everyone knows what it means, and it shouldn't take a link out to explain it.

I've been thinking about this, and I'm inclined to agree with @ybyzek. IMHO the value of presenting the concise answer upfront offsets the additional words.

rmoff:
I'm a hard no on including `quick answer`, and feel strongly we go for just `tl;dr`. 

With both,  we are like dad trying to be cool dancing at the disco, but actually just embarrassing himself. Speaking purely for my own dancing skills here 🕺 

> Reasons to keep (Short answer) are comprehension of term & SEO (as discussed on the call). I lean towards keeping it because not everyone knows what it means, and it shouldn't take a link out to explain it.

Comprehension: we can hyperlink, we can do tooltips; there are plenty of webdev techniques to facilitate explanation if we need it

SEO: I think this is not a sacrifice worth making for SEO. SEO is only part of the requirement; making things that are appealing to developers and speaks to them in their own voice without seeming clunk is another.

ybyzek:
@rmoff if doing both (i.e., `TL;DR (Short Answer)`) is the uncool dad, then let's just make it `Short Answer`.  The concern with having just `TL;DR` is that it leaves folks out of the loop, and we want to be as inclusive as possible so that these tutorials can be accessible and useful for everyone.  `Short Answer` is unambiguous.

ybyzek:
Discussed offline with @rmoff and @gAmUssA -- we agreed to move forward with `Short Answer`

bbejeck (approved):
LGTM

PR #549, opened 2020-09-24
[DO NOT MERGE] CCloud footer

rmoff:
This is a proposal to increase the visibility and awareness of Confluent Cloud in a way that fits more smoothly into tutorials than the existing `Run your app to Confluent Cloud` step, which IMHO doesn't fit logically in the flow of a lot of the tutorials (it's not quite a tutorial, but it's not quite an advert either): 

image

---

IANAWD (I am not a web developer), so the HTML is nasty but this is the _kind_ of thing I am suggesting: 

image

Kapture 2020-09-24 at 16 43 10

---

We could also consider using a similar copy in a more subtle (!) placing further up the tutorial. So long as we are not actually getting in the way of the tutorial (the answer to the problem to which the developer had when they found us from Google), then I think devs accept a little bit of vendor advertising if it's done tactfully.

### DIFF
--- a/_includes/ccloud_banner.html
+++ b/_includes/ccloud_banner.html
@@ -1,0 +1,17 @@
+<div style="background-color: #FFC40C; font-weight: bold; line-height: 1.5;font-size: 12pt; padding: 20px; margin-left: 10%; margin-right: 10%;">
+    <table>
+      <tr>
+        <td>
+          <img src="/assets/img/icon-cloud.svg" width="50"  style="vertical-align: middle;"/>
+  
+        </td>
+        <td style="vertical-align: middle;">
+          Hey - did you know about <a href="https://confluent.cloud">Confluent Cloud</a>? It’s a fully managed Apache Kafka service that also includes managed connectors, ksqlDB, and Schema Registry.
+          <br/><br/>
+        You can <a href="https://confluent.cloud/signup">sign up here</a> to try it out - use the promo code <code>CC100KTS</code> to receive an additional $100 free usage (<a href="https://www.confluent.io/confluent-cloud-promo-disclaimer">T&C</a>). 
+                </td>
+      </tr>
+  
+    </table>
+
+  </div> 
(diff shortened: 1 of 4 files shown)


ybyzek:
@rmoff IAANAWD (I am also not a web developer) but here is what I hacked together.  It keeps the current CCloud section (so that it shows up in the ToC -- which is important for folks who don't naturally scroll to the end) but matches your ideas around color (`assets/sass/tutorial.scss`) and copy (`_includes/shared/markup/ccloud/try-ccloud.adoc`)

Screen Shot 2020-09-24 at 12 35 57 PM

rmoff:
Good point about the ToC. How about this version, which adds in a banner (less obtrusive than the footer) below the ToC and above the tutorial payload? 


Kapture 2020-09-25 at 10 46 19


I do feel that unless the tutorial itself describes running *the actual tutorial* in CCloud then having it as a step in the ToC is misleading. If we actually wrote the bespoke tutorial text per tutorial that showed solving the problem that the tutorial states on CCloud itself it would be ok IMO. 

Thus if we accepted the banner approach below the ToC I'd propose removing the `Deploy on Confluent Cloud
/ Run your app to Confluent Cloud` ToC _generic_ entry.

ybyzek:
@rmoff how about we open this up to the team at the next KT meeting?

rmoff:
SGTM 👍

bbejeck:
FWIW I like the banner approach that sits right below the ToC.

ybyzek:
Notes from meeting today:

- Generally favorable idea with suggestions below
- [Robin AI] do not use yellow, change color to #0074A2 and if possible an 8px corner radius 
- [Robin AI] feedback from Ashlee: remove white lines
- Robin: consider option to include product shot (see top right: https://www.confluent.io/confluent-cloud/tryfree/) -- ping Ashlee if you want to include this
- [Yeva AI] File separate Asana ticket to make top banner more prominent -- DONE

bbejeck:
@rmoff this PR hasn't seen any activity for some time, plus with the changes that offer users the option to view Confluent Cloud directions up-front, I'm going to close this PR.  We can re-open later on if needed

PR #550, opened 2020-09-24
MINOR: Restructure of semaphore yml for running build 

bbejeck:
Testing the restructuring of the semaphore yml file to get builds running under the 50 parallel limit

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -52,9 +52,9 @@ blocks:
             - sem-version java $JAVA_VERSION
             - ./gradlew clean test
 
-  - name: Run the tests
+  - name: Run first block of tests
     execution_time_limit:
-      minutes: 15
+      minutes: 10
     task:
       secrets:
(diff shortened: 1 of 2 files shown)


bbejeck:
updated the yml file and builds trigger automatically again.

Timeout in this PR is dropped to 5 minutes from casual observation, most of the time tests run successfully 2-3 minutes tops.  Anything longer and it most likely going to fail.

WDYT?

\cc @colinhicks @rmoff @ybyzek @gAmUssA @rspurgeon

bbejeck:
Specfically the PR splits the number of tests to run in parallel into blocks of 25 working around the semaphore hard limit of 50 tests per block.

bbejeck:
The `KSQL rekey tests` failed (known flaky test).  Since a test in the first block failed, the second block does not execute and is marked as failed.  We could configure the build to run all blocks regardless of failures in a previous one.

colinhicks (approved):

Issue #510, opened 2020-08-13
[KSQL] rekey example failing after ksqdb version upgrade 


gAmUssA:
during the test via the harness. the output from actual ksql is different from the expected output.
key format should be `KAFKA_INT`

<img width="868" alt="CleanShot 2020-08-13 at 19 27 06@2x" src="https://user-images.githubusercontent.com/433085/90196886-09110080-dd9b-11ea-943b-6d2644de8db1.png">

related #495

gAmUssA:
@colinhicks welcome back. could you take a look on this?

colinhicks:
We had diagnosed this before:
https://github.com/confluentinc/kafka-tutorials/pull/482#issuecomment-663616199

This test doesn't always fail under CI. If data is inserted fast enough, KSQL can infer the more specific key format before the PRINT command is executed. If it's slower, KSQL has less info about the format specificity.

One way to address this is to introduce an optional delay between commands sent by the harness. In this case, this would allow the PRINT command to be deferred so that the reported key format is stable across CI runs.

big-andy-coates:
The ambiguity comes about because the first key(s) successfully deserialize as both `KAFKA_INT` and `KAFKA_STRING`.  When it processes later rows they fail to deserialize as `KAFKA_STRING` and hence we end up with `KAFKA_INT`.

`PRINT` will poll Kafka and iterate over the entire first batch of messages to try and determine the key and value formats.  It then outputs its best guess and the rows in the first batch, before polling again.  If any candidate formats are excluded while processing a batch, `PRINT` will output the new, reduced, candidate list and reprocess the batch.

As Colin alludes to, the timing seems to be the cause of the flakiness here.  It depends what comes back in the first batch as to what the output is.

Taking the example output from https://github.com/confluentinc/kafka-tutorials/pull/482#issuecomment-663156110: 

```
Key format: KAFKA_INT or KAFKA_STRING
Value format: AVRO
rowtime: 2020/07/23 18:00:37.348 Z, key: 780, value: {"RATING": 2.1}
Key format: KAFKA_INT
rowtime: 2020/07/23 18:00:36.359 Z, key: 294, value: {"RATING": 8.2}
rowtime: 2020/07/23 18:00:36.697 Z, key: 354, value: {"RATING": 9.9}
rowtime: 2020/07/23 18:00:37.152 Z, key: 128, value: {"RATING": 8.7}
[...]

PR #555, opened 2020-09-25
Hacky fix to #510 for ksql rekey failing test

rmoff:
See #510 for details of _why_ the test is failing. 

Unless ksqlDB adds support for `SLEEP`, or we split the `docker_ksql_cli_session` and insert a bash `sleep` step (a lot of heavy lifting), this option will hopefully suffice for now. It simply `grep`s out the offending `KAFKA_STRING` mention in the test at the end.


rmoff:
🎉  🎉  🎉  🎉  🎉  🎉  🎉  🎉  

image

colinhicks (approved):
Thanks for tackling this @rmoff! I agree it makes sense to elide the KAFKA_STRING key line if it exists. Maybe we can also add a comment to the Makefile to explain why this is necessary?

bbejeck:
Thanks @rmoff !

PR #565, opened 2020-09-30
GH-186: Adds tutorial for ccloud produce & consume

rspurgeon:

PR #568, opened 2020-09-30
MINOR: Increase timeouts for sempaphore

bbejeck:
Originally I reduced the timeout for semaphore blocks to 5 minutes, but that is too strict as some tests run close to 5 minutes now.  Increasing the timeout to 10 minutes. For example the #565 PR all tests passed, but the build failed because one test ran over 5 minutes.


rspurgeon (approved):
LGTM

PR #575, opened 2020-10-07
Proposed new layout for the front page

rmoff:
Some of the nav is IMHO starting to look a bit messy - we have these category cards and then a lot of words in each one. 

* For example, if you look at the number of occurrences of `console` and `producer` and `consumer` in the `Kafka 101` box, there's a lot of redundancy and thus extra work for people parsing it when figuring out what to click on. 

image

* We should be careful not to bracket _everything_ under 101, since otherwise it just becomes `Kafka` :) What is truly 101? 

I think though that this could improve the nav experience and thus the stickiness of the pages. People hitting this from Google might be fine if they land on a particular tutorial but if we're pushing people to the KT home page to learn stuff we should make it as low-friction as possible to explore

This PR is my proposed amendment to the nav: 

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -33,19 +33,55 @@
         <div class="card">
           <div class="title">
             <img src="{{ "/assets/img/icon-kafka-blue.svg" | relative_url }}" />
-            <div>Kafka 101</div>
+            <div>Producing and Consuming messages</div>
           </div>
           <ul>
-            <li><a href="kafka-console-consumer-producer-basics/kafka.html">Console producer and consumer basics</a></li>
-            <li><a href="ccloud-produce-consume/kafka.html">Console producer and consumer basics with Confluent Cloud</a></li>
-            <li><a href="kafka-console-consumer-primitive-keys-values/kafka.html">Console consumer with primitive keys and values</a></li>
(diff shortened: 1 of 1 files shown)


ybyzek:
@rmoff should this single PR should be split into 2 PRs:

* Copy edit 
* Card number/layout

The reason for decoupling:

1. Card number/layout may require webdev team assistance on developer.confluent.io
2. They are logically differently, and a block on one issue shouldn't necessarily block the other issue

rmoff:
Possibly, yes. We could leave it as-is until if & when webdev do need to get involved? Although I'm not sure they will, if we're just adding a new row?

ybyzek:
Notes from meeting today:

- Generally favorable idea with suggestions below
- [Robin AI]: work with Ashlee to get unique icons for the new cards
- [Robin AI]: new cards require webdev assistance for developer.confluent.io, work with Dan Shafie
- Nice to have: bring up “Be the first to be notified…” to 3rd row
- [Yeva AI]: file GitHub issue to hover over link and show expanded text -- DONE

ybyzek:
@rmoff the new Kafka Tutorial home page has been modified quite a bit since this PR was opened, and some of the concepts proposed in this PR are reflected in the current home page.  If there additional modifications to be made, it will require a new PR.  So closing this one for now.

Issue #570, opened 2020-10-02
How to maintain message order


ybyzek:
Possible discussion points in the KT:

- Ordering per partition
- No ordering across topic
- With idempotency: Idempotent producer (`enable.idempotence=true`)
- Without idempotency:
   - `max.in.flight.requests.per.connection=1` to ensure that only one request can be sent to the broker at a time.
   - To preserve message order while allowing request pipelining: `retries=0` if the application is able to tolerate some message loss.

Note: for idempotent producer, `max.in.flight.requests.per.connection` can be greater than 1 (to allow pipelining) but must 5 or less.  Plus performance numbers: https://cwiki.apache.org/confluence/display/KAFKA/An+analysis+of+the+impact+of+max.in.flight.requests.per.connection+and+acks+on+Producer+performance

Also see notes: 

- https://github.com/apache/kafka/pull/3743
- https://github.com/apache/kafka/blob/2.6/clients/src/main/java/org/apache/kafka/clients/producer/internals/RecordAccumulator.java#L370-L379

ybyzek:
Show seq # in metadata:

```
kafka-run-class kafka.tools.DumpLogSegments \
--print-data-log \ 
--files 00000000000000000000.log \
--deep-iteration
Dumping 00000000000000000000.log
```

ybyzek:
Resolved by https://github.com/confluentinc/kafka-tutorials/pull/587

PR #587, opened 2020-10-12
GH #570 : How to maintain message order and avoid message duplication with the idempotent producer

ybyzek:
This Kafka Tutorial addresses https://github.com/confluentinc/kafka-tutorials/issues/570

Staging: 
http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/GH-570/
http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/GH-570/message-ordering/kafka.html


_includes/tutorials/message-ordering/kafka/markup/dev/make-gradle-wrapper.adoc
```diff
@@ -0,0 +1,5 @@
+Run the following command to obtain the Gradle wrapper:
```
mikebin, line 1: Should a note be added to this tutorial that the user needs to have Gradle installed?

_includes/tutorials/message-ordering/kafka/markup/dev/make-application.adoc
```diff
@@ -0,0 +1,26 @@
...
++++++
+
+The first configuration sets the producer parameter https://kafka.apache.org/documentation/#enable.idempotence[enable.idempotence].
+When set to `true`, it enables an idempotent producer which ensures that exactly one copy of each message is written to the brokers, and in order.
+The default value is `enable.idempotence=false`, so you must explicitly set this to `enable.idempotence=true`.
+For further reading, please see https://cwiki.apache.org/confluence/display/KAFKA/KIP-98+-+Exactly+Once+Delivery+and+Transactional+Messaging#KIP98ExactlyOnceDeliveryandTransactionalMessaging-IdempotentProducerGuarantees[KIP-98].
+
+The second configuration sets the producer parameter https://kafka.apache.org/documentation/#acks[acks].
+The `KafkaProducer` uses the `acks` configuration to tell the leader broker how many acknowledgments to wait for to consider a produce request complete.
+You must set `acks=all` in order to use the idempotent producer, otherwise the producer cannot guarantee idempotence.
```
mikebin, line 14: `KafkaProducer` will automatically set `acks=all` if `enable.idempotence=true`:

```
[main] INFO org.apache.kafka.clients.producer.KafkaProducer - [Producer clientId=myApp] Overriding the default acks to all since idempotence is enabled.
```

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    public Future<RecordMetadata> produce(final String message) {
+        final String[] parts = message.split("-");
+        final String key, value;
+        if (parts.length > 1) {
+            key = parts[0];
+            value = parts[1];
+        } else {
+            key = "NO-KEY";
```
mikebin, line 40: Maybe it's intentional, but `NO-KEY` _is_ a key, and will result in all records being sent to the same partition (vs `null`, which would result in using the uniform sticky partitioner instead)

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
```diff
@@ -0,0 +1,111 @@
...
+        if (args.length < 2) {
+            throw new IllegalArgumentException(
+                    "This program takes two arguments: the path to an environment configuration file and" +
+                            "the path to the file with records to send");
+        }
+
+        final Properties props = KafkaProducerApplication.loadProperties(args[0]);
+
+        props.put(ProducerConfig.ENABLE_IDEMPOTENCE_CONFIG, "true");
+        props.put(ProducerConfig.ACKS_CONFIG, "all");
```
mikebin, line 85: Not required, but doesn't hurt to explicitly set

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
```diff
@@ -0,0 +1,111 @@
...
+            List<Future<RecordMetadata>> metadata = linesToProduce.stream()
+                    .filter(l -> !l.trim().isEmpty())
+                    .map(producerApp::produce)
+                    .collect(Collectors.toList());
+            producerApp.printMetadata(metadata, filePath);
+
+        } catch (IOException e) {
+            System.err.println(String.format("Error reading file %s due to %s", filePath, e));
+        }
+        producerApp.shutdown();
```
mikebin, line 109: Since a shutdown hook is configured, this additional call to `shutdown` isn't needed.

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
```diff
@@ -0,0 +1,111 @@
...
+        final String key, value;
+        if (parts.length > 1) {
+            key = parts[0];
+            value = parts[1];
+        } else {
+            key = "NO-KEY";
+            value = parts[0];
+        }
+        final ProducerRecord<String, String> producerRecord = new ProducerRecord<>(outTopic, key, value);
+        return producer.send(producerRecord);
```
mikebin, line 44: It's probably more common to provide a [callback](https://www.javadoc.io/static/org.apache.kafka/kafka-clients/2.6.0/org/apache/kafka/clients/producer/KafkaProducer.html#send-org.apache.kafka.clients.producer.ProducerRecord-org.apache.kafka.clients.producer.Callback-) to asynchronously handle acks, instead of handling the `Future` returned by `send` directly

_includes/tutorials/message-ordering/kafka/code/build.gradle
```diff
@@ -0,0 +1,63 @@
+buildscript {
+    repositories {
+        jcenter()
+    }
+    dependencies {
+        classpath "com.commercehub.gradle.plugin:gradle-avro-plugin:0.21.0"
```
mikebin, line 6: Is this needed?

_includes/tutorials/message-ordering/kafka/code/build.gradle
```diff
@@ -0,0 +1,63 @@
...
+        classpath "com.github.jengelman.gradle.plugins:shadow:4.0.2"
+    }
+}
+
+plugins {
+    id "java"
+    id "com.google.cloud.tools.jib" version "1.1.1"
```
mikebin, line 13: Is this needed?

_includes/tutorials/message-ordering/kafka/code/build.gradle
```diff
@@ -0,0 +1,63 @@
...
+    maven {
+        url "https://packages.confluent.io/maven"
+    }
+}
+
+apply plugin: "com.commercehub.gradle.plugin.avro"
+apply plugin: "com.github.johnrengelman.shadow"
+
+dependencies {
+    implementation "org.apache.avro:avro:1.10.0"
```
mikebin, line 34: Is this needed?

_includes/tutorials/message-ordering/kafka/code/build.gradle
```diff
@@ -0,0 +1,63 @@
...
+    implementation "org.slf4j:slf4j-simple:1.7.30"
+    implementation "org.apache.kafka:kafka-streams:2.6.0"
```
mikebin, line 36: Perhaps use `kafka-clients` instead of all these Kafka Streams dependencies, since this KT doesn't use Streams?

_includes/tutorials/message-ordering/kafka/code/tutorial-steps/test/dump-log-segments-0.sh
```diff
@@ -0,0 +1,4 @@
+docker-compose exec broker kafka-run-class kafka.tools.DumpLogSegments \
```
mikebin, line 1: Instead of `kafka-run-class kafka.tools.DumpLogSegments`, there's a script `kafka-dump-log` which can be used.

_includes/tutorials/message-ordering/kafka/code/tutorial-steps/test/dump-log-segments-1.sh
```diff
@@ -0,0 +1,4 @@
+docker-compose exec broker kafka-run-class kafka.tools.DumpLogSegments \
```
mikebin, line 1: Instead of `kafka-run-class kafka.tools.DumpLogSegments`, there's a script `kafka-dump-log` which can be used.

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
```diff
@@ -0,0 +1,111 @@
...
+        try {
+            List<String> linesToProduce = Files.readAllLines(Paths.get(filePath));
+            List<Future<RecordMetadata>> metadata = linesToProduce.stream()
+                    .filter(l -> !l.trim().isEmpty())
+                    .map(producerApp::produce)
+                    .collect(Collectors.toList());
+            producerApp.printMetadata(metadata, filePath);
+
+        } catch (IOException e) {
+            System.err.println(String.format("Error reading file %s due to %s", filePath, e));
```
mikebin, line 107: This could be simplified to:

```java
System.err.printf("Error reading file %s due to %s%n", filePath, e);
```

_includes/tutorials/message-ordering/kafka/code/src/test/java/io/confluent/developer/KafkaProducerApplicationTest.java
```diff
@@ -0,0 +1,47 @@
+package io.confluent.developer;
```
mikebin, line 1: Is this intended to show up somewhere in the tutorial? I didn't see it on staging.

_includes/tutorials/message-ordering/kafka/markup/test/consume-log-segments-myTopic-partition-0.adoc
```diff
@@ -0,0 +1,19 @@
...
+
++++++
+<pre class="snippet"><code class="text">{% include_raw tutorials/message-ordering/kafka/code/tutorial-steps/test/expected-log-segments-myTopic-partition-0.txt %}</code></pre>
++++++
+
+Note the familiar `producerId: 0`, which corresponds to the earlier log output when we ran the producer application.
+(If the producer were not idempotent, you would see `producerId: -1`.)
+
+Also observe that each message has a unique sequence number, starting with `sequence: 0` through `sequence: 8`.
+This sequence number is how the broker detects ensures idempotency per partition (if there are retries) and ordered messages (if sequences numbers are sent out of sequence).
```
mikebin, line 18: typo "detects ensures". Also, `retries` must be > 0 if `enable.idempotence=true`

_includes/tutorials/message-ordering/kafka/markup/test/consume-log-segments-myTopic-partition-1.adoc
```diff
@@ -0,0 +1,18 @@
...
+
++++++
+<pre class="snippet"><code class="text">{% include_raw tutorials/message-ordering/kafka/code/tutorial-steps/test/expected-log-segments-myTopic-partition-1.txt %}</code></pre>
++++++
+
+Note the familiar `producerId: 0`, which corresponds to the earlier log output when we ran the producer application.
+(If the producer were not idempotent, you would see `producerId: -1`.)
+
+Also observe that each message has a unique sequence number, starting with `sequence: 0` through `sequence: 2`.
+This sequence number is how the broker detects ensures idempotency per partition (if there are retries) and ordered messages (if sequences numbers are sent out of sequence).
```
mikebin, line 17: typo "detects ensures". Also, `retries` must be > 0 if `enable.idempotence=true`

mikebin (commented):
Overall, looks great. Added a few minor comments.

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
ybyzek: @mikebin good point.  Not intentional in that this code is duplicated from `kafka-producer-application` (cc: @bbejeck ) but in that case, IIUC the partitioner didn't matter.

So stepping back, what is the desirable behavior _for this tutorial_ if there is no key?  I think a true `null` and round robin to use both partitions would better demonstrate ordering on a per-partition basis.  If you agree, I'll modify the code accordingly.  WDYT?

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
ybyzek: @mikebin , sure, I can make this simplification.

@bbejeck this code is also present in `kafka-producer-application` KT and `kafka-producer-application-callback` KT.  Assuming you agree with this simplification, do you have a preference whether this PR addresses those KTs or should I file a separate GH issue?

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
ybyzek: @mikebin , ok, I can validate and remove the call to `shutdown()`.

@bbejeck similar to the previous comment, this code is also present in other KTs. Assuming you agree with this removal, do you have a preference whether this PR addresses those KTs or should I file a separate GH issue?

_includes/tutorials/message-ordering/kafka/code/tutorial-steps/test/dump-log-segments-0.sh
ybyzek: @mikebin I was not aware of that, thank you!  I will validate and use `kafka-dump-log`.

_includes/tutorials/message-ordering/kafka/markup/dev/make-application.adoc
ybyzek: I was aware that setting `enable.idempotence=true` overrides some other producer settings. I was on the fence about whether to include them or not, and I decided to include them for a few reasons:

1. KT can _teach_ users about those settings
2. Sometimes being explicit is better practice because it prevents unintended behavior (user might miss the log message)...
3. Future proof: what if that behavior defaults change in Kafka, etc.

Maybe KT could say that `acks=all` is required for idempotency, so either the user must set it explicitly _or_ whatever setting is there will get overridden by the app.  WDYT?

_includes/tutorials/message-ordering/kafka/markup/dev/make-gradle-wrapper.adoc
ybyzek: Many tutorials use Gradle and I don't think any mention it as a pre-req. In fact, I don't think any KT identifies required local installs to run (in contrast the README mentions pre-reqs for local testing).  Maybe we file a GH issue to track separately?

_includes/tutorials/message-ordering/kafka/markup/dev/make-application.adoc
ybyzek: @mikebin revised copy available, WDYT?

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
ybyzek: @mikebin revised copy available, WDYT?

_includes/tutorials/message-ordering/kafka/markup/test/consume-log-segments-myTopic-partition-1.adoc
ybyzek: @mikebin revised copy available, WDYT?

_includes/tutorials/message-ordering/kafka/markup/dev/make-gradle-wrapper.adoc
mikebin: A separate issue sounds fine to me if any changes are needed at all here.

_includes/tutorials/message-ordering/kafka/markup/dev/make-application.adoc
mikebin: LGTM - thanks!

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
mikebin: `null` makes sense if no key is provided, although in this specific tutorial, keys are provided the input data, so only if the user experiments with their own data would that code be relevant.

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
mikebin: LGTM

ybyzek (commented):
@mikebin thank you for taking the time to review the PR and leaving great comments! 🥇

_includes/tutorials/message-ordering/kafka/markup/test/consume-log-segments-myTopic-partition-1.adoc
mikebin: LGTM

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
ybyzek: I've tested this and it seems that the default partitioner behavior is, as described in https://github.com/apache/kafka/blob/2.6/clients/src/main/java/org/apache/kafka/clients/producer/internals/DefaultPartitioner.java#L30:

> If no partition or key is present choose the sticky partition that changes when the batch is full

Note that it *changes when the batch is full*, and all the null-keyed messages get put into the same batch so they all land in the same partition anyway.  If we wanted to force a different behavior, we could set `batch.size=0` to get more batches and get some kind of distribution, but that conflates different subjects in one KT and could lead to bad copy/paste.

So I'm thinking to change the code as follows:

```
diff --git a/_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java b/_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
index 62a00c7c..45c9fb87 100644
--- a/_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
+++ b/_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
@@ -32,12 +32,12 @@ public class KafkaProducerApplication {
 
     public Future<RecordMetadata> produce(final String message) {
         final String[] parts = message.split("-");
-        final String key, value;
+        String key, value;
+        key = null;
         if (parts.length > 1) {
             key = parts[0];
             value = parts[1];
         } else {
-            key = "NO-KEY";
             value = parts[0];
         }
         final ProducerRecord<String, String> producerRecord = new ProducerRecord<>(outTopic, key, value);
```

...such that a real `null` key gets written, but we just need to understand that it doesn't necessarily mean they will round robin because they will likely land in the same batch.

mikebin: SGTM

abraham-leal:
I noticed there was little mention to the transactional producer. We like to say ordering / idempotency is only guaranteed  for a single producer instance when `enable.idempotence=true` and no transactions are used. I know #570 is abstract enough to maybe include this or not, so I'll leave as a suggestion that a quick shoutout to transactions may be needed to get true idempotency and order across failures / system anomalies.

ybyzek:
@abraham-leal I agree it deserves a shout out, please see https://github.com/confluentinc/kafka-tutorials/pull/587/commits/1eac90b90b32c61fd673ebc2a1979e368f8d4489 which addresses this suggestion.  Thank you.

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
ybyzek: cc: @bbejeck also borrowed from the other KT.  Any leaning on how to fix and where?

_includes/tutorials/message-ordering/kafka/code/build.gradle
ybyzek: @mikebin this line is removed in https://github.com/confluentinc/kafka-tutorials/pull/587/commits/12e0b49c39a1035b172e96a1c1835347b5900090

_includes/tutorials/message-ordering/kafka/code/build.gradle
ybyzek: @mikebin this line is removed in https://github.com/confluentinc/kafka-tutorials/pull/587/commits/12e0b49c39a1035b172e96a1c1835347b5900090

_includes/tutorials/message-ordering/kafka/code/build.gradle
ybyzek: @mikebin this line is removed in https://github.com/confluentinc/kafka-tutorials/pull/587/commits/12e0b49c39a1035b172e96a1c1835347b5900090

_includes/tutorials/message-ordering/kafka/code/build.gradle
ybyzek: @mikebin I did not replace this line in https://github.com/confluentinc/kafka-tutorials/pull/587/commits/12e0b49c39a1035b172e96a1c1835347b5900090 because when I did, it resulted in 

```
java.lang.NoClassDefFoundError: com/fasterxml/jackson/databind/JsonNode
	at org.apache.kafka.clients.producer.internals.TransactionManager.bumpIdempotentEpochAndResetIdIfNeeded(TransactionManager.java:585)
	at org.apache.kafka.clients.producer.internals.Sender.runOnce(Sender.java:311)
	at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:240)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.ClassNotFoundException: com.fasterxml.jackson.databind.JsonNode
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
	... 4 more
```

@gAmUssA had a message on this in Slack, and sounds like it was fixed after 2.6.0 was released.  So I will leave it as-is for now.  I do not want to downgrade to 2.5.0 because it results in different behavior that relevant for this KT:

```
[main] INFO org.apache.kafka.clients.producer.KafkaProducer - [Producer clientId=myApp] Overriding the default retries config to the recommended value of 2147483647 since the idempotent producer is enabled.
```

_includes/tutorials/message-ordering/kafka/code/src/test/java/io/confluent/developer/KafkaProducerApplicationTest.java
ybyzek: Hmmm, not sure about this one.

_includes/tutorials/message-ordering/kafka/code/build.gradle
mikebin: https://issues.apache.org/jira/browse/KAFKA-10378

ybyzek: @mikebin thanks!  @gAmUssA thanks for the workaround.  https://github.com/confluentinc/kafka-tutorials/pull/587/commits/f03c5e4dc979d99bed65cc0477d45ecff9e99b72 should address this now

_includes/tutorials/message-ordering/kafka/markup/dev/make-gradle-wrapper.adoc
ybyzek: Filed https://github.com/confluentinc/kafka-tutorials/issues/591 for tracking

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
bbejeck: @mikebin  good catch

>@bbejeck similar to the previous comment, this code is also present in other KTs. Assuming you agree with this removal, do you have a preference whether this PR addresses those KTs or should I file a separate GH issue?

that was an oversight on my part during refactoring, I'll clean it up in a separate PR

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
bbejeck: same as above I'll put it in the same PR

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
bbejeck: The "NO-KEY" was intentional to provide a default key if a record didn't have one after splitting the line.  Probably poorly worded.

_includes/tutorials/message-ordering/kafka/code/src/test/java/io/confluent/developer/KafkaProducerApplicationTest.java
bbejeck: The project this KT was cloned from has a different test section that uses a unit test.   Part of the build runs only the unit tests, so it might be a good idea to leave it in.

bbejeck (approved):
Nice tutorial @ybyzek LGTM.  I've left some other comments in-line.

bbejeck:
the relevant test for this PR passed

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
ybyzek: Note to self: mirror changes that @bbejeck did in https://github.com/confluentinc/kafka-tutorials/pull/594

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
ybyzek: AI for self: mirror changes that @bbejeck did in https://github.com/confluentinc/kafka-tutorials/pull/594

ybyzek: Addressed by https://github.com/confluentinc/kafka-tutorials/pull/587/commits/44f2b420df1bc802781c0dc7e9c8fde322691f3b

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
ybyzek: Addressed by https://github.com/confluentinc/kafka-tutorials/pull/587/commits/44f2b420df1bc802781c0dc7e9c8fde322691f3b

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
ybyzek: @bbejeck did you have a leaning on whether to leave this as-is, or whether to implement `Callback`?

ybyzek: Thanks for discussing @bbejeck , as we agreed, I will add Callbacks just to this KT and no impact to other KTs

ybyzek: @bbejeck @mikebin https://github.com/confluentinc/kafka-tutorials/pull/587/commits/4214a91f008525b00d78f89479dcca72cfd89fde implements the callback

ybyzek:
@bbejeck @mikebin KT now updated to use Callback.  Please let me know if still good to merge or any suggestions for improvement, thank you.

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
```diff
@@ -0,0 +1,103 @@
...
+        if (parts.length > 1) {
+            key = parts[0];
+            value = parts[1];
+        } else {
+            key = null;
+            value = parts[0];
+        }
+        final ProducerRecord<String, String> producerRecord = new ProducerRecord<>(outTopic, key, value);
+        producer.send(producerRecord,
+            new Callback() {
```
mikebin, line 46: I'd probably write this as a lambda instead of an anonymous inner class (more concise, and implicitly gives you access to the `producerRecord` inside the callback), but not critical.

mikebin (approved):
LGTM - one minor comment, but not a blocker.

bbejeck (approved):
LGTM.

I agree with the comment by @mikebin regarding using a lambda, and I think it should not block the PR either.

_includes/tutorials/message-ordering/kafka/code/src/main/java/io/confluent/developer/KafkaProducerApplication.java
ybyzek: @bbejeck @mikebin very cool suggestion @mikebin .  WDYT about the implementation https://github.com/confluentinc/kafka-tutorials/pull/587/commits/1d8572ba985096b03078355bb948487e8837bb47 ?

mikebin: LGTM!

Issue #562, opened 2020-09-29
Remove config dependency on ZooKeeper (not ZooKeeper the service)


ybyzek:
Instead of configuring a service to connect to ZK, connect to broker.  (This GH issue is NOT to remove ZK from the Docker Compose files).

e.g. replace instances of:

```
SCHEMA_REGISTRY_KAFKASTORE_CONNECTION_URL: 'zookeeper:2181'
```

with

```
SCHEMA_REGISTRY_KAFKASTORE_BOOTSTRAP_SERVERS: 'broker:29092'
```

ybyzek:
Along with the configuration change, there is one more unnecessary ZK dependency, and that is that Schema Registry container should rely just on the broker.  If we could remove the `depends_on` ZooKeeper container (e.g. https://github.com/confluentinc/kafka-tutorials/blob/master/_includes/tutorials/cogrouping-streams/kstreams/code/docker-compose.yml#L43), that would be helpful as well.

ybyzek:
@awalther28 Just as FYI, if https://github.com/confluentinc/kafka-tutorials/pull/579/ weren't already in-flight, this PR could have removed the ZK configuration from the Confluent Metrics Reporter as well (remove `CONFLUENT_METRICS_REPORTER_ZOOKEEPER_CONNECT` in favor of `CONFLUENT_METRICS_REPORTER_BOOTSTRAP_SERVERS`). However, since https://github.com/confluentinc/kafka-tutorials/pull/579/ is already in-flight, that may not be necessary in resolving this PR.

PR #590, opened 2020-10-12
GH #562: Remove config dependency on ZooKeeper (not ZooKeeper the service)

awalther28:
Looking into the single test failure but wanted folks input while investigating. Looks like the same test is consistently failing in master as well 🤔 ([link to master semaphore tests](https://confluentinc.semaphoreci.com/workflows/41706353-512b-4dc2-abf3-a4f987df2513?pipeline_id=61d7652e-19ce-4f9f-be4f-3175cb3070b9))
  
This PR addresses https://github.com/confluentinc/kafka-tutorials/issues/562. It proposes the following:

- remove schema registry's zookeeper dependency in docker-compose
- swap `SCHEMA_REGISTRY_KAFKASTORE_CONNECTION_URL: 'zookeeper:2181'` for `SCHEMA_REGISTRY_KAFKASTORE_BOOTSTRAP_SERVERS: 'PLAINTEXT://broker:9092'`
- alter test for merging/ksql `DESCRIBE EXTENDED` diff check
  - investigation with @bbejeck into the matter, appears changes to configs cause expected behavior (local running stats weren't printed out previously and now they are)


tools/docker-compose.override.yml
```diff
@@ -31,7 +31,6 @@ services:
     environment:
```
ybyzek, line 31: In line 18 above, it should be the case that this line can be removed:

```
      CONTROL_CENTER_ZOOKEEPER_CONNECT: 'zookeeper:2181'
```

gAmUssA:
💯 for removing zookeeper config in SR!

_includes/tutorials/aggregating-average/kstreams/code/docker-compose.yml
```diff
@@ -38,13 +38,12 @@ services:
...
     container_name: schema-registry
     depends_on:
-      - zookeeper
       - broker
     ports:
       - "8081:8081"
     environment:
       SCHEMA_REGISTRY_HOST_NAME: schema-registry
-      SCHEMA_REGISTRY_KAFKASTORE_CONNECTION_URL: 'zookeeper:2181'
+      SCHEMA_REGISTRY_KAFKASTORE_BOOTSTRAP_SERVERS: 'PLAINTEXT://broker:9092'
```
ybyzek, line 46: @awalther28  `PLAINTEXT://` is not required because it reads from `kafkastore.security.protocol` which defaults to PLAINTEXT, https://docs.confluent.io/current/schema-registry/installation/config.html#kafkastore-security-protocol.  If there are multiple SR security types, then this would come into play.  Is the leaning to include `PLAINTEXT://` to be explicit?  The other consideration might be to omit it to lean towards simplicity.

ybyzek (commented):
🏅 @awalther28 great improvement across the board.  Left 2 comments: one missed ZK connect param that can be deleted, and one for consideration.

tools/docker-compose.override.yml
awalther28: Great point, it has been removed in [this commit](https://github.com/confluentinc/kafka-tutorials/pull/590/commits/5d9413a0637aa1ec13ac3439a8c69709ae62257f)

ybyzek:
There is one more tutorial that refers to something ZooKeeper-related:

```
_includes/tutorials/connect-add-key-to-source/kstreams/code/docker-compose.yml:      CONNECT_LOG4J_LOGGERS: org.apache.zookeeper=ERROR,org.I0Itec.zkclient=ERROR,org.reflections=ERROR      
```

Background: I wrote that tutorial and borrowed that configuration line from other GitHub examples.  Admittedly, this was a copy/paste, and so I don't know its purpose or what the impact would be to remove it. 

Do we know if that needs to be there?  Would `org.reflections=ERROR` be sufficient?

_includes/tutorials/aggregating-average/kstreams/code/docker-compose.yml
awalther28: @ybyzek I have a leaning to include `PLAINTEXT://` because I don't think it detracts from simplicity and doesn't rely on default values which can change in future versions (albeit changing default security protocols would be pretty disruptive and unlikely 😉 ). This is not a hill I'm willing to die on. I'm okay with changing this to `'broker:9092'` since that is often how the brokers are referenced in other places throughout the repo.

awalther28: Removed plaintext prefix

ybyzek:
@awalther28 agree, let's leave them as-is.

bbejeck (approved):

bbejeck:
As discussed offline with @awalther28 the failure is unrelated to this PR.

Issue #584, opened 2020-10-08
Address `kafka-streams-schedule-operations` occasional failures


ybyzek:
Context: https://github.com/confluentinc/kafka-tutorials/pull/579

It seems `kafka-streams-schedule-operations` occasionally fails, but these are likely false negatives.  This GitHub issue is to investigate these situations.

bbejeck:
Addressed by #592

PR #592, opened 2020-10-13
MINOR: Revert cp-kafka-connect-base to 5.5.1

bbejeck:
Needed to set `cp-kafka-connect-base` back to `5.5.1` otherwise the tutorial fails due to source topic not getting created.

Covers https://github.com/confluentinc/kafka-tutorials/issues/584

### DIFF
--- a/_includes/tutorials/kafka-streams-schedule-operations/kstreams/code/Dockerfile
+++ b/_includes/tutorials/kafka-streams-schedule-operations/kstreams/code/Dockerfile
@@ -1,4 +1,4 @@
-FROM confluentinc/cp-kafka-connect-base:6.0.0
+FROM confluentinc/cp-kafka-connect-base:5.5.1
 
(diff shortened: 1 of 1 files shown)


ybyzek (approved):
@bbejeck agree to restore functioning, known-working code in the short term to keep the tutorial passing, but do we understand why 5.5.1 passes and 6.0.0 fails?

bbejeck:
>but do we understand why 5.5.1 passes and 6.0.0 fails?

The Kafka Streams app fails because the expected source topic doesn't exist.  So my guess is the connector is failing to start and run but it's building the `datagen` connector locally

```
FROM confluentinc/cp-kafka-connect-base:5.5.1

ENV CONNECT_PLUGIN_PATH="/usr/share/java,/usr/share/confluent-hub-components"

RUN confluent-hub install --no-prompt confluentinc/kafka-connect-datagen:0.4.0
```

At any rate I'll need to look at the log files to see what is happening.

PR #599, opened 2020-10-20
add docker hub login

xli1996:
# what
---------------
Docker hub will have a layer limit about 50/hr pull operation for anonymous users. Add dockerhub cred and dockerhub login to make sure it pull image with confluentsemaphore service account. It should make it have 200/hr pull limit, we could also upgrade the service account to "pro" to have unlimited pulls.

# how
-------------
Add dockerhub cred and docker login.


bbejeck (approved):
LGTM!

bbejeck:
Thanks! @xli1996 !

PR #532, opened 2020-09-08
Move from Sbt to Gradle in produce/consume with scala

DivLoic:
As @gAmUssA requested, this is the Sbt/Gradle conversion for the first produce and consume tutorial. 

[asciicast](https://asciinema.org/a/358242)

Important notes: 
- Two submodules have been created for (produce and consume) to better stick to the jib Gradle plugin
- Dev executions and stdin input didn't work well with Gradle (DEV chapter uses java -jar commands)

PR #612, opened 2020-11-03
fix the produce/consume-scala tutorial by setting up a jar base name

DivLoic:
Here is a modification missing from #532, the migration from sbt to gradle


DivLoic:
Ping @gAmUssA

gAmUssA (approved):
validated locally. will merge to test it on CI

Issue #600, opened 2020-10-22
Provide "Short Answer" for all KT


ybyzek:
This is a follow-on to https://github.com/confluentinc/kafka-tutorials/issues/538

Structural changes to support "Short Answer" provided by PR #540

Now we need to go through all the tutorials and add in short answers ;)

ybyzek:
Structural change is complete, each KT is being ad-hoc upgraded with a short answer

PR #620, opened 2020-11-08
GH-600: add short answer for deserialization-errors KT

ybyzek:
### Description

https://github.com/confluentinc/kafka-tutorials/issues/600

### Staging Docs

http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/GH-600-deserialization-errors/handling-deserialization-errors/ksql.html

### DIFF
--- a/_data/harnesses/deserialization-errors/ksql.yml
+++ b/_data/harnesses/deserialization-errors/ksql.yml
@@ -1,3 +1,11 @@
+answer:
+  steps:
+    - title:
+      content:
+        - action: skip
+          render:
+            file: tutorials/deserialization-errors/ksql/markup/dev/answer-sql.adoc
+
 dev:
   steps:
     - title: Initialize the project
--- a/_data/harnesses/aggregating-count/ksql.yml
+++ b/_data/harnesses/aggregating-count/ksql.yml
@@ -1,6 +1,6 @@
 answer:
   steps:
-    - title: Persistent query using COUNT
+    - title:
       content:
         - action: skip
           render:
--- a/_data/harnesses/aggregating-count/kstreams.yml
+++ b/_data/harnesses/aggregating-count/kstreams.yml
@@ -1,6 +1,6 @@
 answer:
   steps:
-    - title: count() method
+    - title:
       content:
         - action: skip
           render:
--- a/_data/harnesses/console-consumer-producer-basic/kafka.yml
+++ b/_data/harnesses/console-consumer-producer-basic/kafka.yml
@@ -1,11 +1,11 @@
 answer:
   steps:
-    - title: Console producer
+    - title:
       content:
         - action: skip
           render:
             file: tutorials/console-consumer-producer-basic/kafka/markup/answer/produce.adoc
-    - title: Console consumer
+    - title:
       content:
         - action: skip
           render:
--- a/_data/harnesses/masking-data/ksql.yml
+++ b/_data/harnesses/masking-data/ksql.yml
@@ -1,6 +1,6 @@
 answer:
   steps:
-    - title: Use the MASK ksqlDB function
+    - title:
       content:
         - action: skip
           render:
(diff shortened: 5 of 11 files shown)


_data/harnesses/deserialization-errors/ksql.yml
```diff
@@ -1,3 +1,11 @@
+answer:
+  steps:
+    - title:
```
awalther28, line 3: Do we want a title for short answers? We have "short answers" that use the title section (Idempotent producer & Masking Data) and one that doesn't (Anomaly Detection). I personally don't love that the title text appears to be in a larger font size than "Short Answer" and find that a title can be redundant given the nature of this section. Thoughts @ybyzek?

ybyzek: @awalther28

> I personally don't love that the title text appears to be in a larger font size than "Short Answer"

💯 agree, so the omission of a title was intentional.  

For clarity: are you saying you like this approach (as the PR Is now) and we should standardize on this for all KTs?

awalther28: @ybyzek glad to hear I'm not the only one to notice the font size difference. Thank you for looking into if we can change that :) I like this approach of omitting a title and would like to see future KT short answers do the same. 
I could be convinced to standardize adding a title if we can change the font size + the title provides value.

awalther28 (approved):
Overall I approve

_data/harnesses/deserialization-errors/ksql.yml
bbejeck: My preference would be to have no title, so I'd take the PR as it stands now.  But I don't have a strong opinion, so I could go with a title (assuming we work out the font sizing)—just my 2 cents.

bbejeck (approved):
LGTM

_data/harnesses/deserialization-errors/ksql.yml
ybyzek: One more advantage to having no title is that special characters can be added to `adoc` that can't be added to the title.  For example, `KSQL_PROCESSING_LOG` can have backticks, ONLY if it's in adoc, not title

```
`KSQL_PROCESSING_LOG`
```

So the advantages of no title are:

1. Easy, done, no more fiddling with font manipulation
2. Special characters.

Disadvantage:

1. Looks funny?

If we agree with no title, which I think we do...here's an audit:

```
aggregating-count/ksql.yml:answer:
aggregating-count/ksql.yml-  steps:
aggregating-count/ksql.yml-    - title: Persistent query using COUNT
--
aggregating-count/kstreams.yml:answer:
aggregating-count/kstreams.yml-  steps:
aggregating-count/kstreams.yml-    - title: count() method
--
anomaly-detection/ksql.yml:answer:
anomaly-detection/ksql.yml-  steps:
anomaly-detection/ksql.yml-    - title:
--
console-consumer-producer-basic/kafka.yml:answer:
console-consumer-producer-basic/kafka.yml-  steps:
console-consumer-producer-basic/kafka.yml-    - title: Console producer
--
deserialization-errors/ksql.yml:answer:
deserialization-errors/ksql.yml-  steps:
deserialization-errors/ksql.yml-    - title:
--
masking-data/ksql.yml:answer:
masking-data/ksql.yml-  steps:
masking-data/ksql.yml-    - title: Use the MASK ksqlDB function
--
message-ordering/kafka.yml:answer:
message-ordering/kafka.yml-  steps:
message-ordering/kafka.yml-    - title: Configure the producer for idempotency
```

So I could update this PR to fix _all_ the KTs that use titles, and move it to adocs.  @awalther28 @bbejeck  please let me know if you agree, and I'll jump on it ;)

awalther28: I agree 💯

bbejeck: +1 for me

PR #622, opened 2020-11-09
MINOR: update gem install command

bbejeck:
### Description
The build was failing to compile the jekyll site due to `invalid option: --no-ri --no-doc` failure

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -33,7 +33,7 @@ blocks:
           - checkout
           - cache restore
           - npm install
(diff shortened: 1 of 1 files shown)


colinhicks (approved):
LGTM. I guess that the `gem` executable version changed on Semaphore, although I didn't see it noted in their changelog.

bbejeck:
> I guess that the gem executable version changed on Semaphore, although I didn't see it noted in their changelog.

Yeah, that's the only thing that made sense to me, but I couldn't find where it would have changed. 
Thanks for the review

PR #644, opened 2020-11-23
MINOR_remove_ZK_CONNECTION_URL

ybyzek:
### Description

Replace link from ZK to broker.
High confidence this should pass, but let's wait for tests to complete before merging.

cc @dlfaraujo as FYI


bbejeck (approved):
LGTM!

PR #645, opened 2020-11-24
MINOR: Remove invalid gem options

bbejeck:
### Description

When I fixed the build in #622 I neglected to do the same update for the live-site deploy yml file.


colinhicks (approved):

bbejeck:
thanks @colinhicks

PR #648, opened 2020-11-30
DEVX-2288: KT Anomaly Detection test is flaky

awalther28:
### Description
Add sleep between populating table and stream. There seem to be some other tests failing in block 1, so I moved this test up to block 1 to ensure it got executed with every semaphore run. Over 6 valid runs (valid run=block is run where this test was present) in semaphore, this test passed every time. Locally this is running without problems for me. 

### Staging Docs

http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/devx-2288


bbejeck (approved):
LGTM

PR #683, opened 2021-01-11
MINOR: Update KT index page 

bbejeck:
### Description

PR for adding proposed changes to KT index page

This PR incorporates the work from @rmoff in  #575 and suggestions from @ybyzek 

### Staging Docs

http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/MINOR_update_index_page_layout


bbejeck:
This is an initial PR for breaking up the `Apply a function to data`.  It will be updated soon to incorporate the changes proposed in #575

ybyzek (commented):
@bbejeck this looks so much better, love it!

Some suggested tweaks:

1. Some cards have a period, some do not.  They should all be consistent
2. My leaning is that 3 cards wide looks better than 4 cards wide -- but this is very subjective and you can ignore if you prefer 4
3. How about instead of `Create X windows`, simplify to `X windows`?
4. Does `Find distinct events` belong in `Routing Events`, or is there a better card for it?

colinhicks:
> My leaning is that 3 cards wide looks better than 4 cards wide -- but this is very subjective and you can ignore if you prefer 4

A related consideration is that the CSS is written with 3 cards in mind. As implemented, the 4-card layout appears to work at viewport width of ~1450px and above. Between 1080px and 1450px, the outer card content is cut off. 

It would be possible to use breakpoints to switch between 3 and 4 cards depending on browser widths, but this would require some changes to the styles and markup. If part of the goal is optimization for screen widths > ~1450, this effort is probably reasonable overall. If not, I suggest sticking with 3 cards.

bbejeck:
Updates

1. Only 3 cards per row
2. Shortened titles to fit on one line, no-breaks
3. Address comments from @ybyzek https://github.com/confluentinc/kafka-tutorials/pull/683#pullrequestreview-578311188

index.html
```diff
@@ -27,81 +27,106 @@ <h2 class="subtitle">Apache Kafka is a powerful, scalable, fault-tolerant distri
...
             <img src="{{ "/assets/img/icon-kafka-blue.svg" | relative_url }}" />
-            <div>Kafka 101</div>
+            <div>Producing and Consuming</div>
+          </div>
+           <ul>
+            <li><a href="kafka-console-consumer-producer-basics/kafka.html">The basics</a></li>
+            <li><a href="kafka-console-consumer-primitive-keys-values/kafka.html">Primitive keys and values</a></li>
+            <li><a href="kafka-console-consumer-read-specific-offsets-partitions/kafka.html">Reading from an offset and partition</a></li>
+            <li><a href="kafka-producer-callback-application/kafka.html">Using Callbacks</a></li>
+            <li><a href="ccloud-produce-consume/kafka.html">Using Confluent Cloud</a></li>
```
ybyzek, line 45:
```suggestion
            <li><a href="ccloud-produce-consume/kafka.html">Using Confluent Cloud CLI</a></li>
```

ybyzek:
@bbejeck please note https://github.com/confluentinc/kafka-tutorials/pull/697 should be reviewed/merged into this PR before merging (it fixes the resolution of the new images)

Also, looks like this tutorial is missing https://kafka-tutorials.confluent.io/message-ordering/kafka.html .  I compared `release` to this PR, and it looks like it's only one off, so hopefully this is the only missing one:

```
grep html index.html | grep li | wc -l 
```

ybyzek (changes requested):
Marking note on required change but overall looks really good :)

ybyzek (approved):
@bbejeck 🥳 LGTM

bbejeck:
Updated for comments.

I took a stab at setting the max height of the card and adding scrolling.  Check out the `Manipulating Streams` card as that has overflow content. 

\cc @colinhicks for CSS changes

bbejeck:
I think we need a way to make it obvious when users need to scroll.

Ideas anyone?  \cc @colinhicks

ybyzek:
@bbejeck sorry, missed this in my last review, I think latest commit is still missing:

```
          <li><a href="message-ordering/kafka.html">Idempotent producer for message ordering and no duplication</a></li>		
```

Once that is added, LGTM.

bbejeck:
@ybyzek it's there I didn't use the correct link text - updating now

ybyzek:
Thank you @bbejeck ! 👍

bbejeck:
Thanks for the reviews @ybyzek and @colinhicks !

PR #685, opened 2021-01-20
MINOR: Changes for using vault for secrets

bbejeck:
### Description
Changes for using vault to get credentials vs semaphore secrets


time-less-ness (approved):
Eeeyep! Looks good.

bbejeck:
one test failure, `KSQL finding distinct events`, unrelated to changes

bbejeck:
kicked off another round of tests

colinhicks (changes requested):
great stuff @bbejeck. Since this project uses other pipelines based on the `promotions` config, and these pipelines use the `aws_credentials` secret, I think we will need to configure each additional pipeline for Vault as well:
- live-site-deploy.yml
- pr-staging-deploy.yml
- staging-site-deploy.yml 

And looks like the config there can be a bit simpler:

```
global_job_config:
  secrets:
    - name: vault_sem2_approle_prod
  prologue:
    commands:
      - checkout
      - make install-vault
      - . vault-bin/vault-setup
      - . vault-sem-get-secret aws_credentials
```

.semaphore/semaphore.yml
```diff
@@ -61,18 +77,6 @@ blocks:
     execution_time_limit:
       minutes: 10
     task:
-      secrets:
-        - name: aws_credentials
-        - name: dockerhub-semaphore-cred
```
bbejeck, line 66: >And looks like the config there can be a bit simpler:

I'm not sure as we needed to pull the `dockerhub-semaphore-cred` with semaphore secrets from before.

bbejeck:
> I think we will need to configure each additional pipeline for Vault as well:

@colinhicks I'm not sure, I think it's taken care of by the global config.  The [pr-staging-deploy.yml](https://confluentinc.semaphoreci.com/jobs/54473b19-83b4-4ce4-89f0-ab406263315a) successfully deployed with the current configuration 

>And looks like the config there can be a bit simpler:

I'm not sure about this one either, we needed to pull

.semaphore/semaphore.yml
colinhicks: I was suggesting a separate config. The tasks in those pipelines just do stuff with S3. They don't pull private images.

colinhicks:
@bbejeck Yeah, I'm not 100% certain, but I think that the global_job_config is global within a given pipeline, and promotions are essentially downstream pipelines, which I am assuming do not inherit the config of the pipeline from which they were triggered. Here are the docs I referenced: https://docs.semaphoreci.com/reference/pipeline-yaml-reference/#global_job_config

> The pr-staging-deploy.yml successfully deployed with the current configuration

Since the existing secrets have not been removed, this pipeline could continue to work using the credentials provided by the existing secret. One way to narrow this down is by removing this from the `*-deploy.yml` configs:
```
secrets:	
        - name: aws_credentials
```

bbejeck:
> Since the existing secrets have not been removed, this pipeline could continue to work using the credentials provided by the existing secret

DOH! yeah you're right about that, I overlooked that those are separate config files.  Yeah, they need to use vault as well.  I'll update the PR

bbejeck:
@colinhicks updated the PR for the pipeline configs to use vault as well.  Good catch on that!

bbejeck:
With the last PR, test run a test timed out, but the `pr-staging-deploy` using [vault succeeded ](https://confluentinc.semaphoreci.com/jobs/3374fc65-ef51-4283-80a2-3018d4409dd8)

colinhicks (approved):
LGTM. Thanks for tackling this, @bbejeck!

PR #696, opened 2021-01-28
update ksql version update script to work with linux

agavra:
### Description

The `update-ksqldb-version.sh` script now works for both linux and darwin based OS.

### Staging Docs

n/a

### Testing Done

Runs on semaphore :) https://confluentinc.semaphoreci.com/branches/abe8dbf4-589f-41ac-bb5a-4665fbf40983

### DIFF
--- a/tools/update-ksqldb-version.sh
+++ b/tools/update-ksqldb-version.sh
@@ -24,5 +24,10 @@ fi
 cd "$(dirname $0)/.."
 
 # both the xargs and sed incanations may need to be changed for non-macOS contexts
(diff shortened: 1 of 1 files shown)


colinhicks (approved):
Thanks @agavra! LGTM.

Btw, I think this is a GNU sed difference, so if a Mac OS user has manually installed the GNU flavor, the wrong branch would be chosen here. Seems like a minor edge case though, and not something that needs to be solved for here.

agavra:
nice catch @colinhicks - I suppose this isn't a regression though, so we can cross that bridge when we get there :P

PR #703, opened 2021-02-04
chore: divide Semaphore jobs into two execution blocks

colinhicks:
This will roughly double time in CI, hopefully with the benefit of reduced flakiness.


agavra (approved):
thanks!

cprasad1:
lgtm

PR #781, opened 2021-04-02
Fix ksqlDB finding distinct tutorial for new key syntax and default retention

vcrfxia:
### Description

The "finding distinct events" tutorial is broken for ksqlDB versions 0.15+ due to changes in multi-column group by behavior. This PR updates the statements to use the JSON key format to fix the issue, and also sets a long retention so the windowed table is not empty.


lct45 (approved):
@vcrfxia thanks for catching this, LGTM!

PR #799, opened 2021-04-17
MINOR:Update the vault app role name

bbejeck:
### Description
Make required updates for semaphore access to vault


awalther28 (approved):
The tests get kicked off and the staging site is being deployed as expected. It looks like the KT issues with semaphore will be resolved with the merging of this PR. @bbejeck do you agree?

bbejeck:
Thanks @awalther28 I always forget about the other yml files

PR #806, opened 2021-04-21
CCloudify finding distinct events

awalther28:
### Description

Add a "Confluent" section for the distinct events KT. 

### Staging Docs

http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/convert-distinct/finding-distinct-events/confluent.html


bbejeck (approved):
Nice work @awalther28, LGTM.  As a sanity check, I'd like to see the test pass for this tutorial before merging.

awalther28:
Thanks for the review @bbejeck, the latest semaphore run triggered the test and the test did pass.

ybyzek:
@awalther28 awesome job!  It is helpful to manually validate that the steps work (both the app and the `ccloud` commands).  You probably already did this, but just wanted to check, have you done that manual validation once for this KT?  (The down side of not having the validation in place for CCloud KTs -- need at least one manual validation)

awalther28:
@ybyzek yes I have manually validated the Confluent workflow but did not walk through the Kafka Streams workflow. The one thing I didn't love is that the Confluent consumer output used escape characters in the URL string `({"ip":"10.0.0.2","url":"https:\/\/www.confluent.io\/hub\/confluentinc\/kafka-connect-datagen","timestamp":"2019-09-16T14:53:43+00:01"})` but the KT output doesn't show those escape characters nor does Confluent Cloud. The explanation text above the consumer output says something like "you should see messages similar to the following" which is technically accurate 🤷‍♀️ I can update the output with escape characters if you think that's better!
I'll manually validate the kafka streams workflow.

awalther28:
@ybyzek Still no real feedback from the CLI team about the escape characters 🙁

PR #853, opened 2021-05-17
CCloudify kafka producer callback

awalther28:
### Description

Convert kafka producer callback. Manually stepped through the Confluent steps. I have verified that the test [passes in semaphore.
](https://confluentinc.semaphoreci.com/jobs/34a8033e-f463-4deb-aafb-8be2d72bc71a)
Added a short answer. 

Removed unnecessary `input.topic.name` config value from property files. 

### Staging Docs
http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/convert-kafka-callback/kafka-producer-callback-application/confluent.html

### New tutorial checklist

- [x] Add a `Short Answer`


_includes/tutorials/kafka-producer-application-callback/confluent/markup/dev/make-dev-file.adoc
```diff
@@ -0,0 +1,21 @@
+Then create a development file at `configuration/dev.properties`:
+
++++++
+<pre class="snippet"><code class="shell">{% include_raw tutorials/kafka-producer-application-callback/confluent/code/configuration/dev.properties %}</code></pre>
++++++
+
+Let's do a quick walkthrough of some of the producer properties.
```
ybyzek, line 7: @awalther28 If it's the case that everything below is the same as the non-CCloud tutorial, can you dedupe (either with an include or a separate adoc file?

ybyzek (changes requested):
@awalther28 great conversion!

1. Where does `replication.factor=3` get set in this tutorial, or does it not need it if using the default broker RF?   (IIRC producer can use default broker RF, whereas it was Kafka Streams that doesn't support that yet)
2. Also note misspelling in summary: `handing` instead of `handling` (not by this PR, but in original KT commit).  If you don't mind fixing here...?

```
you want to have an automatic way of handing the responses after producing records
```

_includes/tutorials/kafka-producer-application-callback/confluent/markup/dev/make-dev-file.adoc
awalther28: Yea totally, I split it up into a new doc and added an include for that new doc. I verified the rendering locally. Work was done in commit https://github.com/confluentinc/kafka-tutorials/pull/853/commits/5c580ee2d045b60b81a8f05a59eb6a1ed235ab9f

awalther28:
@ybyzek to answer your questions: 
> Where does replication.factor=3 get set in this tutorial, or does it not need it if using the default broker RF? (IIRC producer can use default broker RF, whereas it was Kafka Streams that doesn't support that yet)

-- It doesn't need to be set because we aren't using the admin client or a streams app to create the topic. Here we are using cloud cli to create the topic. 

> Also note misspelling in summary: handing instead of handling (not by this PR, but in original KT commit). If you don't mind fixing here...?

-- Yes fixed in commit https://github.com/confluentinc/kafka-tutorials/pull/853/commits/5c580ee2d045b60b81a8f05a59eb6a1ed235ab9f !

ybyzek (approved):
Ah right @awalther28 . Thanks!

LGTM, 🚢

PR #912, opened 2021-08-02
Fixed typo exctractor->extractor

javabrett:


ybyzek (approved):
@javabrett good catch, LGTM

PR #913, opened 2021-08-05
Update class names for font awesome icons

danshafie:
### Description

Fixing social icon issues that was reported here: [https://confluent.slack.com/archives/C19MZFS9W/p1628121074006400](url)

Font awesome is included on the website, we just need to update the class names.

also added transition so when user hovers they know its clickable.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -59,27 +59,27 @@
         <ul>
           <li>
             <a target="_blank" href="https://twitter.com/ConfluentInc">
-              <i class="fa fa-twitter"></i>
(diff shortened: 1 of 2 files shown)


ybyzek (approved):
@danshafie thanks for the PR.  Staging site isn't built by Semaphore because it's a fork but will assume you have made the correct changes.

danshafie:
> @danshafie thanks for the PR. Staging site isn't built by Semaphore because it's a fork but will assume you have made the correct changes.

yeah, it worked correctly when built locally

Issue #921, opened 2021-08-20
Clean up old pages from S3


ybyzek:
https://kafka-tutorials.confluent.io/kafka-connect-datagen-ccloud/kafka.html had been removed a few months back but it is still a live page.  I suspect this is because this page persists in S3 and so it keeps getting loaded.  We need to clean out the S3 bucket and ensure only intended pages are discoverable.

ybyzek:
Solutions to investigate (thanks @colinhicks !)

1. This needs to be changed to use sync https://github.com/confluentinc/kafka-tutorials/blob/master/.semaphore/live-site-deploy.yml#L24 (https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html)
2. https://docs.aws.amazon.com/AmazonS3/latest/userguide/CustomErrorDocSupport.html

PR #1002, opened 2021-12-10
Migrate to /tutorials (take 3)

ybyzek:
### Staging Docs

http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/tutorials


.semaphore/live-site-deploy.yml
```diff
@@ -20,5 +20,5 @@ blocks:
             - npm install
             - gem install bundler
             - bundle install
-            - JEKYLL_ENV=production bundle exec jekyll build
-            - aws s3 cp --recursive ./_site s3://kafka-tutorials.confluent.io/
+            - JEKYLL_ENV=production bundle exec jekyll build --baseurl /tutorials
+            - aws s3 cp --recursive ./_site s3://kafka-tutorials.confluent.io/tutorials
```
hartantothio, line 24: Looking at this, looks like you guys are uploading the built site to `kafka-tutorials.confluent.io/tutorials`.

Am I correct to assume that whatever is already in S3 on the root, will continue to stay there? For example, if there is a file in `kafka-tutorials.confluent.io/robots.txt`, it will stay there?

ybyzek: @hartantothio cc: @bbejeck Due to https://github.com/confluentinc/kafka-tutorials/issues/921, existing assets do not get deleted from S3.  As for updating it, the new builds would all put `robots.txt` at kafka-tutorials.confluent.io/robots.txt (see http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/tutorials/robots.txt as evidence).  Is this a problem?

ybyzek:
Folks, we are about to kick off the series of steps for PRs and Cloudfront changes that will result in KT migration to CD.

@hartantothio is merging https://github.com/confluentinc/web/pull/1418, and after that we need this PR to push through -- it moves assets in S3 to `/tutorials`

Looking for final approval of this PR:

- Required: @bbejeck @hartantothio (@colinhicks -- if you have the time)
- Optional: @rspurgeon 

After which we will need to follow quickly with a second PR to merge `master` into `release.

Please let me know if you have any concerns, Thanks!

_includes/nav.html
```diff
@@ -8,15 +8,7 @@
 
     <!-- Logo(s) -->
     <a class="logo-container" href="{{ "/" | relative_url }}">
```
hartantothio, line 10: I think this needs to be hardcoded to just be `/`. Right now, when I mouse hover logo, it takes me to `/tutorials`. Ideally, we want to take user to homepage (which would be confluent developer).

hartantothio (approved):
Looks good overall aside from the url for the logo

_includes/nav.html
ybyzek: @hartantothio fixed in latest commit: https://github.com/confluentinc/kafka-tutorials/pull/1002/commits/6ad758342feb5c8674d0b2b549d84e4430d6047c

ybyzek:
> This set of changes looks good to me.
> 
> Is there any period of time where the content would be available at both the old and new URLs? If so it may be worth adding a rel=canonical tag to the pages.

I'll defer to @hartantothio on this, but IIUC he is going to add 301 redirects such that a user going to the old URL will be automatically redirected to the new URL, so the period of time should be very small.  Assuming this to be the case, are canonical tags still worth the effort?

colinhicks:
got it. If the redirects are going in around the same time as the URL cut-over, I would think canonical tags are not worthwhile.

PR #1021, opened 2022-01-20
Skip confluent projects from build

bbejeck:


bbejeck:
@ybyzek this change will automatically skip `confluent` projects from the java test run.  We probably should have put this in before any way.

ybyzek (approved):
@bbejeck thanks for identifying this and fixing!

PR #1029, opened 2022-01-24
Snippet width fix

scott-confluent:


ybyzek (approved):
Validated locally, thanks @scott-confluent

PR #1061, opened 2022-02-02
feat: [UIF-1175] - Remove Kafka-tutorials: Update swiftype site ids

mt206:
### Description
https://confluentinc.atlassian.net/browse/UIF-1175

### DIFF
--- a/index.html
+++ b/index.html
@@ -8,7 +8,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="description" content="{{ site.description }}">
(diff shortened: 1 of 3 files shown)


ybyzek (commented):
@mt206 thanks for the PR.  I'm not familiar with these tags so who would be a required approver?

joel-hamill (approved):

mt206:
> @mt206 thanks for the PR. I'm not familiar with these tags so who would be a required approver?

@ybyzek Hi. There isn't a specific person on the kafka-tutorials side that I know of but Joel is familar with the project and with the previous changes made so his sign-off should be good.

PR #1092, opened 2022-02-25
Breadcrumbs on the tutorial and use-case pages

bpurkaystha:
### Description

[This](https://app.asana.com/0/1201817272699202/1201822439776478) is the Asana story. In short, this PR is about adding two breadcrumbs: `Back to tutorials` on each tutorial & use-case page.

Screen Shot 2022-02-25 at 10 08 45 PM

### DIFF
--- a/_layouts/recipe.html
+++ b/_layouts/recipe.html
@@ -35,7 +35,7 @@
     <div class="hero-body">
       <div class="container">
         <ul class="breadcrumb">
-          <li><a href="/tutorials/use-cases">< Back to tutorials</a></li>
+          <li><a href="/tutorials/use-cases">&lt; Back to tutorials</a></li>
         </ul>
         {% if page.community_contribution %}
         <span class="community-contribution">Community contribution ✨</span>
--- a/_layouts/recipe.html
+++ b/_layouts/recipe.html
@@ -31,10 +31,12 @@
 </head>
 <body class="pg-tutorial">
   {% include nav.html %}
-
   <section class="hero">
     <div class="hero-body">
       <div class="container">
+        <ul class="breadcrumb">
+          <li><a href="/tutorials/use-cases">< Back to tutorials</a></li>
+        </ul>
         {% if page.community_contribution %}
         <span class="community-contribution">Community contribution ✨</span>
         {% endif %}
@@ -122,6 +124,14 @@
       {% include contribute-tutorial-lang.html %}
       {% else %}
       {% include tutorial-content.html %}
+      <section class="section">
+        <div class="container">
+          <ul class="breadcrumb">
+            <li><a href="/tutorials/use-cases">&lt; Back to tutorials</a></li>
+          </ul>
+        </div>
+
+      </section>
       {% endif %}
     {% endif %}
   </div>
--- a/_layouts/recipe.html
+++ b/_layouts/recipe.html
@@ -130,7 +130,6 @@
             <li><a href="/tutorials/use-cases.html">&lt; Back to tutorials</a></li>
           </ul>
         </div>
-
       </section>
       {% endif %}
     {% endif %}
--- a/_layouts/recipe.html
+++ b/_layouts/recipe.html
@@ -35,7 +35,7 @@
     <div class="hero-body">
       <div class="container">
         <ul class="breadcrumb">
-          <li><a href="/tutorials/use-cases">&lt; Back to tutorials</a></li>
+          <li><a href="/tutorials/use-cases.html">&lt; Back to tutorials</a></li>
         </ul>
         {% if page.community_contribution %}
         <span class="community-contribution">Community contribution ✨</span>
@@ -127,7 +127,7 @@
       <section class="section">
         <div class="container">
           <ul class="breadcrumb">
-            <li><a href="/tutorials/use-cases">&lt; Back to tutorials</a></li>
+            <li><a href="/tutorials/use-cases.html">&lt; Back to tutorials</a></li>
           </ul>
         </div>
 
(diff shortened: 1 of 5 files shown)


ybyzek (commented):
Thanks for the PR @bpurkaystha .  The recipes need breadcrumbs too, not just tutorials (those linked from http://127.0.0.1:4000/use-cases.html)

_layouts/recipe.html
```diff
@@ -31,10 +31,12 @@
 </head>
 <body class="pg-tutorial">
   {% include nav.html %}
-
   <section class="hero">
     <div class="hero-body">
       <div class="container">
+        <ul class="breadcrumb">
+          <li><a href="/tutorials/use-cases">&lt; Back to tutorials</a></li>
```
ybyzek, line 38: Should this be use-cases.html?  (Note addition of `.html`)

ybyzek (changes requested):

_layouts/recipe.html
bpurkaystha: Updated.!

ybyzek (approved):
Validated and LGTM.  @danshafie @bbejeck -- did you want to take a look at this PR before we merge?

assets/sass/main.scss
```diff
@@ -404,6 +404,26 @@ a {
   }
 }
 
+.breadcrumb {
+  a {
+    color: #edeaeafa !important;
```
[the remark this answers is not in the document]
bpurkaystha: I'm in process of talking to Ashlee regarding this..!

_layouts/recipe.html
```diff
@@ -122,6 +124,14 @@ <h5 class="title is-5">Video walkthrough</h5>
       {% include contribute-tutorial-lang.html %}
       {% else %}
       {% include tutorial-content.html %}
+      <section class="section">
+        <div class="container">
+          <ul class="breadcrumb">
+            <li><a href="/tutorials/use-cases.html">&lt; Back to tutorials</a></li>
+          </ul>
+        </div>
+
```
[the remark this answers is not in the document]
bpurkaystha: Removed.

_layouts/tutorial.html
```diff
@@ -260,6 +263,14 @@ <h5 class="title is-5">Video walkthrough</h5>
       {% include contribute-tutorial-lang.html %}
       {% else %}
       {% include tutorial-content.html %}
+      <section class="section">
+        <div class="container">
+          <ul class="breadcrumb">
+            <li><a href="/tutorials/">&lt; Back to tutorials</a></li>
+          </ul>
+        </div>
+
```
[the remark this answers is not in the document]
bpurkaystha: Removed.

assets/sass/main.scss
```diff
@@ -404,6 +404,26 @@ a {
...
+  li {
+    margin-left: 0;
+  }
+}
+
+.section {
+  .container {
+    .breadcrumb {
+      a {
+        color: #000000 !important;
```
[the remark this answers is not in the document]
bpurkaystha: Updated...!

assets/sass/main.scss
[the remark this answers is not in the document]
bpurkaystha: For now, I've started using `$color_ltGray` from the `_vars.scss` file.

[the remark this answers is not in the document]
bpurkaystha: Actually, Ashlee thinks it should be white. Therefore, now it's changed to using $color_white

danshafie:
thank you @bpurkaystha

danshafie (approved):

package-lock.json
```diff
@@ -1,23 +1,8 @@
 {
   "name": "kafka-tutorials",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
ybyzek, line 4: @bpurkaystha @danshafie I might have missed a comment about this, but just checking that the change to this file is intended.

bpurkaystha: It appears that I've had a different node version when installing the packages, but that should not be too much of a concern. Thanks @ybyzek

ybyzek: Since this isn't required for the PR, can you please back out these changes?

bpurkaystha: Actually, in my opinion, this should be present there in the PR. Ideally, we'd want to ensure possiblity of time-travel of the past states. That's why we should keep it included in the PR. Moreover, this lock file contributes to the performance improvement by reducing the need to read `package.json` itself. It should be there in the version control.

ybyzek: @bpurkaystha if the file should be updated, can we have this PR just about the breadcrumb and you can submit a separate PR for `package-lock.json`?

bpurkaystha: Updated..!

colinhicks: I agree the lockfile change isn't a big deal. Looks like the one in master was generated by a newer version of npm (>= 7) compared to the one in the referenced commit.

Since the project doesn't specify a semver for the npm version in the package.json this sort of change is expected and should be harmless. I think the only side effect is churn in this file. If one PR changes the lockfileVersion to `1`, and a developer pulls the changes, then runs `npm install` with a newer version of npm, there would be new changes to the lockfile. If those changes were committed, then pulled to an environment with an older npm version, running `npm install` would flip the version back. And so forth.

For this project and its relatively light set of node dependencies this will probably never matter, but if you want to avoid the flip-flopping, you can set something like `"engines": { "npm": ">=7.0.0"}` in the package.json. This should require an npm that uses the new lockfile version and avoid the flip-flop.

ybyzek: @colinhicks https://github.com/confluentinc/kafka-tutorials/issues/1093 tracks your comment

@bpurkaystha thanks for backing out the change

PR #1110, opened 2022-03-03
GH #1013 Use count instead of having

### LINKED ISSUE #1013: Update ksqlDB finding distinct tutorial
The https://kafka-tutorials.confluent.io/finding-distinct-events/ksql.html uses a table with a windowed aggregation, but to provide distinct events it needs to create an additional stream to filter out the results with tombstones due to the semantics of using a table. 

But the tutorial as is a little confusing in its current form.  We should update the first query to remove the `HAVING` and add a `COUNT` and then update the second query to filter on `count == `

bbejeck:
### Description

Fixes #1013 
### Staging Docs

 http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/GH-1013/finding-distinct-events/ksql.html


mjsax (approved):
Only skimmed over it briefly. LGTM.

ybyzek (approved):
@bbejeck LGTM

PR #1215, opened 2022-03-25
Recipes: Fix query typo in online-dating final answer

colinhicks:


ybyzek (approved):
Good catch, thanks for the PR @colinhicks

PR #1228, opened 2022-04-06
update ksqldb-latest in line with master

swist:
### Description
This branch takes `master` then merges back changes from `ksqldb-latest` back in


colinhicks (commented):
Thanks for doing this! It seems like we need more intervention if we take this route. I noticed that the merge yielded an [invalid semaphore yaml](https://confluentinc.semaphoreci.com/workflows/df96023b-84b5-4a37-a495-f56d589c2907?pipeline_id=582e3464-7f36-48d9-9c79-ac6400224e8b). And the complexity of the merge history means that necessary changes from master were dropped, like: https://github.com/confluentinc/kafka-tutorials/commit/6d7bbb86c705266882ccd0d425e3ea946ef6afd9.

I wonder if it would be easier to do the following:
* PR into master to update the semaphore config so that jfrog-based images can be used in CI
* After merging that PR, branch off master to update all the docker-compose files to the RC under test (using this [util](https://github.com/confluentinc/kafka-tutorials/blob/master/tools/update-ksqldb-version.sh)).
* Validate the CI build for the new branch.
* Force push the new branch as ksqldb-latest.

PR #1239, opened 2022-04-14
cleaning up tutorial hero: moving launch button and breadcrumb

scott-confluent:
### Description

Per discussions, we're cleaning up the hero and making the "Launch..." button more prominent.

Screen Shot 2022-04-14 at 1 26 39 PM

### DIFF
--- a/_includes/recipe-content-answer.html
+++ b/_includes/recipe-content-answer.html
@@ -1,29 +1,21 @@
-<div>
+<div class="launch-container">
 
    {% if site.data.harnesses[page.static_data][page.stack].answer %}
 
-   <br>
-   <br>
-   <br>
-   <br>
-
-     <div class="container" style="background-color: #E6F5FB; padding: 30px; ">
-       <div class="columns">
-         <div class="column is-full">
-           <h3 class="title">Launch recipe in Confluent Cloud</h3>
            {% for step in site.data.harnesses[page.static_data][page.stack].answer.steps %}
 
             <div class="tutorial-try-it-step content">
+            {% if step.title %}
              <h4 class="subtitle" id="{{ step.title | slugify }}">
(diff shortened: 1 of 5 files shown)


_includes/recipe-content-answer.html
```diff
@@ -1,39 +1,28 @@
...
-       <div class="columns">
-         <div class="column is-full">
-           <h3 class="title">Launch recipe in Confluent Cloud</h3>
            {% for step in site.data.harnesses[page.static_data][page.stack].answer.steps %}
 
             <div class="tutorial-try-it-step content">
+            {% if step.title %}
              <h4 class="subtitle" id="{{ step.title | slugify }}">
-               <div class="text">{{ step.title }}</div>
+               <div class="text">{{ step.title }} Fake Title</div>
```
iPzard, line 10: Is this ok to keep "Fake Title"?

assets/sass/main.scss
```diff
@@ -256,6 +256,14 @@ a {
     }
   }
 
+  .container {
+    margin: 15px 0;
+
+    .title {
+      color: #E6F5FB;
```
iPzard, line 263: No variable to use for this value?

iPzard (approved):

ybyzek (changes requested):
Thanks for the PR.  Some comments

1. Although this `index.html` page isn't used on the live site, the hero centering has changed:

Screen Shot 2022-04-14 at 16 16 32 PM

2. Now there is a mismatch: the non-recipe tutorials have `Back to recipes` in the hero, but the recipes have that `Back to recipes` under the hero.  I also think it's weird having it below.  Can we move it back to where it was, because it looks better (this is subjective and just my opinion) and is consistent with the other pages?

ybyzek (approved):
Thanks for the fixes, LGTM!

PR #1255, opened 2022-05-02
adding swiftype meta tags for tutorials and recipes

scott-confluent:
### Description

The swiftype tags are missing for tutorials and recipes. Adding those and a `category` tag for the new developer search https://app.asana.com/0/1178383736931373/1156641992335726

### DIFF
--- a/_layouts/recipe.html
+++ b/_layouts/recipe.html
@@ -8,6 +8,9 @@
 
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1" />
+  <meta class="swiftype" name="site-id" data-type="integer" content="3" />
+  <meta class="swiftype" name="docs-boost" data-type="integer" content="6" />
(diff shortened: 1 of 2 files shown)


_layouts/recipe.html
```diff
@@ -8,6 +8,9 @@
 
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1" />
+  <meta class="swiftype" name="site-id" data-type="integer" content="3" />
+  <meta class="swiftype" name="docs-boost" data-type="integer" content="6" />
+  <meta class="swiftype" name="category" data-type="string" content="recipe" />
```
ybyzek, line 13: Does it make sense to differentiate `recipe` from `tutorial`?  How is it consumed?

scott-confluent: I suppose that's more of a question for the content creators? My plan was to distinguish them on the code side, but if a user selects `tutorial` it would include both

ybyzek (approved):
LGTM

PR #1267, opened 2022-05-11
Remove threads and use MockConsumer.schedulePollTask

bbejeck:
### Description

The test uses threads which leads to flaky failures due to race conditions.  To fix this, the test should use the `MockConsumer.schedulePollTask`instead.  

### Staging Docs

This PR includes a testing change only, so no need to view deployed tutorial.


bbejeck:
The test [for this PR](https://confluentinc.semaphoreci.com/jobs/550ecb80-65ab-4ed4-b434-f25a994aa211) passed

ybyzek (approved):
@bbejeck LGTM

PR #1271, opened 2022-05-23
Add copy with instructions for recipes

ybyzek:
### Description

Be more instructive with new copy for recipes

### Staging Docs

e.g. http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/add-instr/customer-journey/confluent.html

### DIFF
--- a/_layouts/recipe.html
+++ b/_layouts/recipe.html
@@ -63,7 +63,10 @@
       {% unless page.help_wanted %}
         <div class="example">
           <div class="example-content">
-            <div class="text">{{ site.data.tutorials[page.static_data].introduction }}</div>
+            <div class="text">
+               {{ site.data.tutorials[page.static_data].introduction }}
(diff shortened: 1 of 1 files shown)


bbejeck (approved):

bbejeck:
All tests passed! Merge it now!

colinhicks:
Have we considered a different visual treatment for this paragraph? Since this new text is repeated with the same content on all recipes, I wonder if it would be better if it were pulled into a call-out. 

image

ybyzek:
@colinhicks I personally don't like the text as much when it's got the box around it, but happy to consider if you feel strongly about it.  I suppose it's personal preference. 

In terms of existing pattern, we have `short answer` format that potentially could be reused (see https://developer.confluent.io/tutorials/kafka-console-consumer-producer-basics/confluent.html for an example)

colinhicks:
I don't feel strongly. Just wanted to talk it through briefly. I think the short answer styles may present too much of a visual speed bump. Let's go with the proposal as is :)

colinhicks (approved):

Issue #1268, opened 2022-05-14
Topic name issue


marcin2x4:
Where does 'ORDERS_ENRICHED' topic gets created?
I checked the ksql code but couldn't find in provided [files](https://github.com/confluentinc/kafka-tutorials/blob/master/_includes/tutorials/denormalization/confluent/code/tutorial-steps/dev/source.sql).

https://github.com/confluentinc/kafka-tutorials/blob/2cacde636a6036b80efd51c67ceea56167fa41b0/_includes/tutorials/denormalization/confluent/code/tutorial-steps/dev/sink.sql#L7

ybyzek:
@marcin2x4 thanks for filing the issue. This may be a valid bug (IIRC the stream was renamed but the topic referenced in the connector was not).

PR #1274, opened 2022-05-27
update ksql with sales_orders_enriched

danicafine:
### Description

Solves https://github.com/confluentinc/kafka-tutorials/issues/1268.

Added topic name for sales_orders_enriched stream, updated references, and made other topic names more consistent.


daveklein (approved):
LGTM

ybyzek (changes requested):
@danicafine thanks for the PR.

`VALUE_FORMAT` and `PARTITIONS` are not required -- they will inherit from the parent streams.  For consistency with the other recipes, these parameters should be removed from the `sales_orders_enriched` creation

ybyzek (approved):
LGTM!  Thanks @danicafine

PR #1277, opened 2022-05-31
Rename kafkacat to kcat

ybyzek:
### Description

Rename kafkacat to kcat, per product name change
https://github.com/edenhill/kcat

### Staging Docs

http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/kcat/how-to-count-messages-on-a-kafka-topic/kafka.html

### DIFF
--- a/_includes/tutorials/connect-add-key-to-source/kafka/code/tutorial-steps/dev/consume-topic.sh
+++ b/_includes/tutorials/connect-add-key-to-source/kafka/code/tutorial-steps/dev/consume-topic.sh
@@ -1,3 +1,3 @@
-docker exec -i kafkacat kafkacat -b broker:9092 -t postgres_cities \
+docker exec -i kcat kcat -b broker:9092 -t postgres_cities \
             -C -s avro -r http://schema-registry:8081 -e \
             -f 'Key     (%K bytes):\t%k\nPayload (%S bytes):\t%s\n--\n'
--- a/_data/harnesses/count-messages/kafka.yml
+++ b/_data/harnesses/count-messages/kafka.yml
@@ -32,7 +32,7 @@ dev:
           render:
             file: tutorials/count-messages/kafka/markup/dev/02c-wait-for-containers.adoc
 
-    - title: Run kafkacat to count the messages
+    - title: Run kcat to count the messages
       content:
         - action: execute
           file: tutorial-steps/dev/03a-count-messages.sh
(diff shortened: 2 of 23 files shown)


_includes/tutorials/connect-add-key-to-source/kafka/code/tutorial-steps/dev/consume-topic.sh
```diff
@@ -1,3 +1,3 @@
-docker exec -i kafkacat kafkacat -b broker:9092 -t postgres_cities \
+docker exec -i kcat kcat -b broker:9092 -t postgres_cities \
```
edenhill, line 1: I think we removed the superfluous `kcat` after the image name when we renamed the confluentinc/kafkacat->kcat docker image (and it was never there for the edenhill docker images).

So, all docker commands that use the edenhill/kcat image, or the newer confluentinc/kcat image, should remove the extra  `kcat` argument

edenhill (changes requested):

_includes/tutorials/connect-add-key-to-source/kafka/code/tutorial-steps/dev/consume-topic.sh
ybyzek: @edenhill the KT usage of `kcat` (and `kafkcat` before) seems to have been overriding the `entrypoint`, so the extra `kcat` is required.

```
  kcat:
    image: edenhill/kcat:1.7.1
    container_name: kcat
    links:
      - broker
    entrypoint:
      - /bin/sh
      - -c
      - |
        apk add jq;
        while [ 1 -eq 1 ];do sleep 60;done
```

So removing the extra `kcat` results in an error:

```
docker-compose exec kcat -b broker:9092 -t postgres_cities \     
            -C -s avro -r http://schema-registry:8081 -e \
            -f 'Key     (%K bytes):\t%k\nPayload (%S bytes):\t%s\n--\n'

OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "-b": executable file not found in $PATH: unknown
```

ybyzek:
> I don't fancy overriding the entry-point since these examples will seem to be using the standard upstream edenhill/kcat image, but in practice gives incompatible examples that will infact not work with that vanilla image.

@edenhill agree, and thank you for raising this issue.  I don't know the logic for why the example was written this way...will need to investigate separately.  We are decoupling the rename of the tool (this PR) from the logic of how to call it (tracking in #1278 ).

PR #1279, opened 2022-06-06
Revert change mainClass.set back to mainClassName

bbejeck:


bbejeck:
change addressed a deprecation we need to keep it

PR #1305, opened 2022-07-18
updating swiftype tags -- category -> contentType

scott-confluent:
### Description

We had an issue on .io where we needed to change from `category` to `contentType` and it makes sense to keep everything aligned.


davetroiano (approved):
This looks good.  no staging site generated since it's a fork but I validated nothing breaks by manually editing developer.confluent.io
image

scott-confluent:
@davetroiano thanks!

Issue #1326, opened 2022-07-27
build automation shouldn't install harness_runner as root


davetroiano:
The semaphore build installs harness-runner as root (system-wide). It should use a more sandboxed approach.

davetroiano:
related to #1324

Issue #1324, opened 2022-07-27
build failure: rm: cannot remove 'kafka-tutorials/harness_runner/harness_runner.egg-info/top_level.txt': Permission denied


davetroiano:
Semaphore build sometimes hits this error:

![image](https://user-images.githubusercontent.com/4550245/181285092-a92bcb38-bd18-43d4-a7c0-4a5139e8aa30.png)

ybyzek:
@davetroiano I don't recall seeing this error before.  I could have just missed it but is this a new issue (cc: @bbejeck ) and if so, why is it surfacing now?

davetroiano:
Won't fix, instead fixing #1326 which makes this moot

PR #1327, opened 2022-07-27
change harness_runner egg permissions to non-root in semaphore build

davetroiano:
This mitigates https://github.com/confluentinc/kafka-tutorials/issues/1324 and is
a short-term fix until https://github.com/confluentinc/kafka-tutorials/issues/1326
is taken on.

### Description

https://github.com/confluentinc/kafka-tutorials/issues/1324

### Staging Docs

N/A

### New tutorial checklist

N/A

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -21,6 +21,7 @@ global_job_config:
       - docker login --username $DOCKERHUB_USER --password $DOCKERHUB_APIKEY
       - docker login --username $DOCKER_USER --password $DOCKER_APIKEY confluent-docker.jfrog.io
       - sudo pip3 install -e harness_runner/
+      - sudo chown -R semaphore:semaphore harness_runner/ # mitigation for https://github.com/confluentinc/kafka-tutorials/issues/1324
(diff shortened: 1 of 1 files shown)


ybyzek:
@davetroiano I meant to leave a comment here instead of the GH issue, but just curious and I'll repost here: https://github.com/confluentinc/kafka-tutorials/issues/1324#issuecomment-1198262249

bbejeck:
> > @davetroiano I meant to leave a comment here instead of the GH issue, but just curious and I'll repost here: [#1324 (comment)](https://github.com/confluentinc/kafka-tutorials/issues/1324#issuecomment-1198262249)
> 
> It started cropping up after moving to self-hosted agents. Admittedly that's not root cause, but my theory is that it's due to a difference in user / default perms / pip behavior on non-self-hosted vs self-hosted agents.

@davetroiano @ybyzek - I agree with it being related to the move to self-hosting.

davetroiano:
This is no longer needed with the better fix in https://github.com/confluentinc/kafka-tutorials/pull/1331

Issue #1328, opened 2022-07-27
payment-status-check semaphore build flake


davetroiano:
I have observed a few instances of this test failing:

```
make -C _includes/tutorials/payment-status-check/ksql-test/code tutorialFailed in 
01:21
make: Entering directory '/home/semaphore/kafka-tutorials/_includes/tutorials/payment-status-check/ksql-test/code'00:00
rm -r tutorial-steps/test/outputs || true00:00

...

# Check presence of entry01:21
grep "NOT OK" tutorial-steps/test/outputs/validate/output-0.log || { cat tutorial-steps/test/outputs/validate/output-0.log ; false; }01:21
+---------------------------------------------------------------------------+---------------------------------------------------------------------------+---------------------------------------------------------------------------+01:21
|PAYMENT_ID                                                                 |STATUS_COUNTS                                                              |SERVICE_STATUS_LIST                                                        |01:21
+---------------------------------------------------------------------------+---------------------------------------------------------------------------+---------------------------------------------------------------------------+01:21
|1                                                                          |{OK=2}                                                                     |[{ "system" : "AML", "status" : "OK"}, { "system" : "FUNDS", "status" : "OK|01:21
|                                                                           |                                                                           |"}]                                                                        |01:21
|15                                                                         |{OK=1}                                                                     |[{ "system" : "AML", "status" : "OK"}]                                     |01:21
|9                                                                          |{OK=2}                                                                     |[{ "system" : "AML", "status" : "OK"}, { "system" : "FUNDS", "status" : "OK|01:21
|                                                                           |                                                                           |"}]                                                                        |01:21
|7                                                                          |{OK=2}                                                                     |[{ "system" : "AML", "status" : "OK"}, { "system" : "FUNDS", "status" : "OK|01:21
|                                                                           |                                                                           |"}]                                                                        |01:21
|17                                                                         |{OK=1}                                                                     |[{ "system" : "AML", "status" : "OK"}]                                     |01:21
|3                                                                          |{OK=2}                                                                     |[{ "system" : "AML", "status" : "OK"}, { "system" : "FUNDS", "status" : "OK|01:21
|                                                                           |                                                                           |"}]                                                                        |01:21
|11                                                                         |{OK=1}                                                                     |[{ "system" : "AML", "status" : "OK"}]                                     |01:21
[...]

PR #1329, opened 2022-07-27
add a wait before validating ksqlDB recipe tests relying on table output

davetroiano:
The handful of tests that query table output tend to be flaky due to the
async nature of ksqlDB. There's no guarantee that table result are
available after corresponding INSERTS. Other recipes that validate
via a stream query like "SELECT ... EMIT CHANGES LIMIT n;" aren't
susceptible to the same test flakiness since they will block on
n rows being emitted..

### Description

https://github.com/confluentinc/kafka-tutorials/issues/1328

### Staging Docs

N/A

### New tutorial checklist

N/A

### DIFF
--- a/_data/harnesses/logistics/ksql-test.yml
+++ b/_data/harnesses/logistics/ksql-test.yml
@@ -44,4 +44,20 @@ test:
             skip: true
           stdin:
             - file: ../../confluent/code/tutorial-steps/dev/manual.sql
+
+        - name: wait for table to be populated
+          action: sleep
+          ms: 5000
+          render:
+            skip: true
+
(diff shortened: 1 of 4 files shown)


davetroiano:
one data point in favor of sleep mitigation: I am able to repro #1328 locally (on 2 of 6 runs) and with this change I haven't hit it in 6 runs

bbejeck (approved):
Thanks for the PR @davetroiano 

LGTM pending tests passing - hopefully, this will fix the flaky tests

davetroiano:
I stopped build after recipe tests passed

PR #1331, opened 2022-07-28
non-root harness-runner install

davetroiano:
### Description

https://github.com/confluentinc/kafka-tutorials/issues/1326

### Staging Docs

N/A

### New tutorial checklist

N/A


bbejeck:
This is looking better @davetroiano! 
The one failure on the first build seems unrelated, second build almost got an all green - seems like the spurious hanging we get with certain tests is the culprit, but again, it seems unrelated

bbejeck (approved):
LGTM - IMHO enough of the build ran to signal success

PR #1339, opened 2022-08-03
Added click tracking

hartantothio:
### Description
Send click event (link clicks) to Segment.

### Staging Docs

http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/add-click-tracking


assets/js/segment.js
```diff
@@ -0,0 +1,79 @@
...
+
+      if (!payload) {
+        // Do nothing
+        return;
+      }
+
+      // Traverse the locations
+      const location = path.reduce((location, node) => {
+        // Using "data-tracking-location"
+        if (!node?.dataset?.trackingLocation) {
```
mt206, line 64: Interesting. I didn't know optional chaining and destructuring was safe to use nowadays without a transpiler but it looks like it is supported in all major browsers.

assets/js/segment.js
```diff
@@ -0,0 +1,79 @@
...
+      text = target.textContent;
+    }
+
+    return {
+      hrefUrl,
+      text: (text || '').trim(),
+    };
+  }
+
+  document.body.addEventListener('click', function (e) {
```
mt206, line 49: Are you tracking all clicks, regardless of whether or not it is a link? If you track clicks through bubbling then this will also capture clicks on blank sections of the page. Not sure if that matters or not

assets/js/segment.js
hartantothio: We should be fine -- I looked at our visitor traffic and compared that to supported browsers.

assets/js/segment.js
hartantothio: Yes, using bubbling, we can track all clicks regardless of DOM hierarchy. I do have additional checks on whether or not to send the events.

hartantothio:
@bbejeck @davetroiano - would you be able to help in reviewing this PR?

hartantothio:
@bbejeck @davetroiano - would you be able to help review?

I'm also asking somebody from my team to review this. Thanks.

TerryDevS (approved):
LGTM

bbejeck (approved):

PR #1340, opened 2022-08-04
update anomaly detection tutorial to use timestamps based on run time

davetroiano:
Updates the anomaly detection tutorial so that the automated `ksq-test-runner` test uses hard-coded values, while the records in markup have dynamic recent (within the past seven days) timestamps so that the tutorial feels fresher and so that the non-`ksql-test-runner` part of build automation isn't subject to race condition (old records getting deleted).


_includes/tutorials/anomaly-detection/ksql/code/tutorial-steps/dev/print-accounts-to-monitor.sql
```diff
@@ -1,4 +1,4 @@
-SELECT USERNAME, WINDOW_START, WINDOW_END
```
davetroiano, line 1: weakness

davetroiano:
push pipeline tests pass

bbejeck (approved):
LGTM!

PR #1341, opened 2022-08-04
MINOR: Replace deprectated 'TIMESTAMPTOSTRING' with 'FORMAT_TIMESTAMP'

bbejeck:
### Description

Updated insert statements to use 'UNIX_TIMESTAMP()' to generate dates

This PR proposes to solve the "too old" date issue by using `UNIX_TIMESTAMP()` to generate dates for inserting the initial data for the tutorial.  Additionally, the tutorial uses a deprecated method `TIMESTAMPTOSTRING`, so it's been replaced with the recommended`FORMAT_TIMESTAMP` function.


_includes/tutorials/session-windows/ksql/code/tutorial-steps/dev/populate-stream.sql
```diff
@@ -1,9 +1,10 @@
...
-INSERT INTO clicks (ip, timestamp, url) VALUES ('51.56.119.117','2019-07-18T10:01:36Z','/convallis/nunc/proin.jsp');
-INSERT INTO clicks (ip, timestamp, url) VALUES ('53.170.33.192','2019-07-18T10:02:00Z','/vestibulum/vestibulum/ante/ipsum/primis/in.json');
-INSERT INTO clicks (ip, timestamp, url) VALUES ('51.56.119.117','2019-07-18T11:03:21Z','/vehicula/consequat/morbi/a/ipsum/integer/a.jpg');
-INSERT INTO clicks (ip, timestamp, url) VALUES ('51.56.119.117','2019-07-18T11:03:50Z','/pede/venenatis.jsp');
-INSERT INTO clicks (ip, timestamp, url) VALUES ('53.170.33.192','2019-07-18T11:40:00Z','/nec/euismod/scelerisque/quam.xml');
-INSERT INTO clicks (ip, timestamp, url) VALUES ('53.170.33.192','2019-07-18T11:40:09Z','/ligula/nec/sem/duis.jsp');
+INSERT INTO clicks (ip, timestamp, url) VALUES ('51.56.119.117',FORMAT_TIMESTAMP(FROM_UNIXTIME(UNIX_TIMESTAMP()),'yyyy-MM-dd''T''HH:mm:ssX'),'/etiam/justo/etiam/pretium/iaculis.xml');
+INSERT INTO clicks (ip, timestamp, url) VALUES ('51.56.119.117',FORMAT_TIMESTAMP(FROM_UNIXTIME(UNIX_TIMESTAMP() + (60 * 1000)),'yyyy-MM-dd''T''HH:mm:ssX'),'/nullam/orci/pede/venenatis.json');
```
bbejeck, line 2: The tutorial needs timestamps at different intervals to create sessions.  The idea is to add time in milliseconds to the current time - here the format is `Seconds * 1000`

_includes/tutorials/session-windows/ksql/code/tutorial-steps/dev/populate-stream.sql
```diff
@@ -1,9 +1,10 @@
...
+INSERT INTO clicks (ip, timestamp, url) VALUES ('53.170.33.192',FORMAT_TIMESTAMP(FROM_UNIXTIME(UNIX_TIMESTAMP() + (91 * 1000)),'yyyy-MM-dd''T''HH:mm:ssX'),'/mauris/morbi/non.jpg');
+INSERT INTO clicks (ip, timestamp, url) VALUES ('51.56.119.117',FORMAT_TIMESTAMP(FROM_UNIXTIME(UNIX_TIMESTAMP() + (96 * 1000)),'yyyy-MM-dd''T''HH:mm:ssX'),'/convallis/nunc/proin.jsp');
+INSERT INTO clicks (ip, timestamp, url) VALUES ('53.170.33.192',FORMAT_TIMESTAMP(FROM_UNIXTIME(UNIX_TIMESTAMP() + (2 * 60 * 1000)),'yyyy-MM-dd''T''HH:mm:ssX'),'/vestibulum/vestibulum/ante/ipsum/primis/in.json');
```
bbejeck, line 5: (Minutes * Sixty Seconds * 1000)

davetroiano (changes requested):
Should we do anything about stale dates with the `ksql-test-runner`? i.e., this part [here](https://developer.confluent.io/tutorials/create-session-windows/ksql.html#test-it)

I ran into this [here](https://github.com/confluentinc/kafka-tutorials/pull/1340)

Maybe the compromise is to let the `ksql-test-runner` / "Test it" section have stale dates (maybe update them to be August 2022 rather than July 2019 as part of this, but leave them hard coded?), but also add infinite retention to topics [here](https://github.com/confluentinc/kafka-tutorials/pull/1340) to prevent race condition build failures due to topic cleanup due to default retention period not being large enough.

davetroiano:
oops I missed this:
>NOTE There's one issue remaining - the test data relies on hardcoded dates in a JSON file, so the tests for this PR will fail until we develop approach for creating test data

How about:
1. update the "dev" queries and expected outputs to not select timestamp / window boundaries
2. update the "test" input / output timestamps to be hardcoded but recent
3. add infinite retention so that we're safe against race condition build failures

That would get us to a happy place of improved quality and reliable build, right?

bbejeck:
@davetroiano  - PR updated 

>Should we do anything about stale dates with the ksql-test-runner? i.e., this part [here](https://developer.confluent.io/tutorials/create-session-windows/ksql.html#test-it)

After spending a good deal of time on this, I think we should leave the dates for the test runner alone because updating requires a lot of manual work for little-to-no gain.

The timestamps used in the tutorial are set to use the current time so it should always work as expected for users working through the tutorial

settings.gradle
```diff
@@ -25,7 +25,7 @@ def preDir = {
     return SKIP_SUBTREE
   }
 }
-
+// Dummy comment to force build
```
davetroiano, line 28: remove

bbejeck:
Only one test failure and it's unrelated

Test for this PR passed https://confluentinc.semaphoreci.com/jobs/d8c26926-d73d-4f55-acfe-0feb12b2628f

davetroiano (approved):
LGTM! just remove the dummy comment before merging

davetroiano:
>After spending a good deal of time on this, I think we should leave the dates for the test runner alone because updating requires a lot of manual work for little-to-no gain.

@bbejeck sounds good, I agree

settings.gradle
bbejeck: ah yeah good catch

bbejeck:
latest change unrelated to files for the tutorial so merging now

bbejeck:
Thanks for the review @davetroiano!

PR #1350, opened 2022-08-19
Parallel Consumer tutorial

davetroiano:
### Description

[asana](https://app.asana.com/0/1202377393885559/1202924894981919/f)

### Staging Docs

http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/parallel-consumer-tutorial/

### New tutorial checklist

 - [x] Implement good test cases (if relevant)
 - [x] Validate hyperlinks
 - [x] Spell check (e.g. using `aspell`)
 - [x] Full code validated in Confluent Cloud
 - [x] SQL style/syntax consistent to existing recipes
 - [x] Validate presentation with `bundle exec jekyll serve --livereload`
 - [x] Tutorial added to appropriate `index.html` or `use-case.html` file


bbejeck:
The build is failing because this line is missing from `settings.gradle`
```
include 'includes:tutorials:kafka-parallel-consumer-application:kafka:code'
```

bbejeck:
Great tutorial, @davetroiano! I've made a pass over it - overall looks good.

davetroiano:
> Great tutorial, @davetroiano! I've made a pass over it - overall looks good.

thanks for taking a pass @bbejeck! i'll incorporate your feedback and make a few more usability improvements before considering this ready to merge

where do you think this should live on the [tutorials home page](https://developer.confluent.io/tutorials/)?  I didn't put too much thought into what I went with here.  I could see `Learn the basics` > `Build applications` > `Confluent Parallel Consumer`, or a new `Confluent Parallel Consumer` section under `Master advanced concepts` that would initially contain just this Parallel Consumer hello world, if we add other PC tutorials (for vert.x / reactor, more realistic applications / use cases), they'd go in this section. Maybe to start we do the first, and consider separating it out only if / when we hit 2 tutorials. Lt me know what you think

chuck-confluent:
REVIEW IN PROGRESS-- I'm gonna come back to this

## General comments
- I actually like building up a local directory from scratch rather than git cloning at the beginning. Maybe though there should be a link to the completed solution in github

## Specific comments

### Write the cluster information into a local file
- [ ] " From the Clients view, get the connection information customized to your cluster" -- I think the full path is Data Integration -> Clients -> New client -> Java.
- [ ] "Create new credentials for your Kafka cluster and Schema Registry" -- I might suggest writing a description like "parallel-consumer-tutorial" so the keys are easy to find and delete later

### Create a topic
- [ ] I think the topic name should be less generic. Maybe prefixed by "parallel-consumer" to avoid name collisions. "parallel-consumer.input" or something

### Configure the project
- [ ] can we at least use java 11? Java 8 is getting long in the tooth. It doesn't really matter functionally in this case, but it might be a signal to devs that we're not cool. Maybe. I don't know.
- [ ] This is the first time I'm seeing the word Gradle. Should probably list dependencies up at the very start of the tutorial. I would suggest a prerequisites section, including links to help folks install gradle and the confluent cli and anything else required. It turns out to be a pain in the butt to install the latest version of gradle. I tried to install with apt and it gave me v4.4(suuuper old!!!!). I tried to install with snap and it turns out WSL doesn't mess with snap. I don't have brew, but that would probably be easiest. This turned out to be a lot of friction, so we should make it easier for folks. Maybe include a little script like
```bash
wget -c https://services.gradle.org/distributions/gradle-7.5.1-bin.zip -P /tmp
sudo unzip /tmp/gradle-7.5.1-bin.zip -d /opt/gradle
sudo chmod +x /opt/gradle/gradle-7.5.1/bin/gradle 
sudo ln -s /opt/gradle/gradle-7.5.1/bin/gradle /usr/bin/gradle
````
Ideally this would all be preconfigured in a 1-click lab environment, but that's out of scope of this review.

- [ ] "And be sure to run the following command to obtain the Gradle wrapper" -- Maybe just a quick heads up what the wrapper does and why it's necessary.

### Create the Confluent Parallel Consumer Application

File: _includes/tutorials/confluent-parallel-consumer-application/kafka/code/src/main/java/io/confluent/developer/PropertiesUtil.java [hunk 1]
```diff
@@ -0,0 +1,22 @@
...
+ * Utility class for loading Properties from a file.
+ */
+public class PropertiesUtil {
+
+  public static Properties loadProperties(String fileName) throws IOException {
+    final Properties props = new Properties();
+    final FileInputStream input = new FileInputStream(fileName);
+    props.load(input);
+    input.close();
+    return props;
```
bbejeck, line 19: We could use a `try-with-resources` block here.  But I'd say handle that in a follow-up PR

_includes/tutorials/confluent-parallel-consumer-application/kafka/code/src/main/java/io/confluent/developer/SleepingRecordHandler.java
```diff
@@ -0,0 +1,27 @@
...
+  public SleepingRecordHandler(final int numMilliseconds) {
+    this.numMilliseconds = numMilliseconds;
+  }
+
+  @Override
+  protected void processRecordImpl(final ConsumerRecord consumerRecord) {
+    try {
+      Thread.sleep(numMilliseconds);
+    } catch (InterruptedException e) {
+      throw new RuntimeException(e);
```
bbejeck, line 23:
```suggestion
      Thread.currentThread().interrupt();
```

bbejeck (approved):
Really nice job on this tutorial @davetroiano!

I have a couple of minor comments - but I don't think they should block the PR - we can do them in a follow-up one

_includes/tutorials/confluent-parallel-consumer-application/kafka/code/src/main/java/io/confluent/developer/PropertiesUtil.java
bbejeck: @davetroiano I resolved this by accident - it's still relevant

bbejeck:
> While the performance test runs, take a few sips of the beverage…​ actually never mind. It’s done:

I loved this line in the tutorial!

_includes/tutorials/confluent-parallel-consumer-application/kafka/code/src/main/java/io/confluent/developer/SleepingRecordHandler.java
davetroiano: 👍  made this change

_includes/tutorials/confluent-parallel-consumer-application/kafka/code/src/main/java/io/confluent/developer/PropertiesUtil.java
davetroiano: done

PR #1355, opened 2022-08-25
update deployment instructions

davetroiano:
* example release PR description so releasers can copy/paste
* change instructions to requiring squash and merge since that's the
  only option allowed


bbejeck (approved):
Thanks, @davetroiano LGTM

PR #1365, opened 2022-09-01
Updating tutorials to use new shadowJar location.

daveklein:
The location of the shadowJar Gradle plugin has moved, so I updated the pertinent line in the `build.gradle` file for all the tutorials that use shadowJar.  28 files were affected.

### DIFF
--- a/_includes/tutorials/aggregating-average/kstreams/code/build.gradle
+++ b/_includes/tutorials/aggregating-average/kstreams/code/build.gradle
@@ -4,7 +4,7 @@ buildscript {
   }
   dependencies {
     classpath "com.commercehub.gradle.plugin:gradle-avro-plugin:0.22.0"
-    classpath "com.github.jengelman.gradle.plugins:shadow:6.1.0"
+    classpath "gradle.plugin.com.github.jengelman.gradle.plugins:shadow:7.0.0"
   }
 }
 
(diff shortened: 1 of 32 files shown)


davetroiano (commented):
thanks @daveklein!

Not a code change, but need to address with Dev Productivity: upgrade gradle on self hosted agents to get past automated build error
```
> Failed to apply plugin class 'com.github.jengelman.gradle.plugins.shadow.ShadowBasePlugin'.00:20
   > This version of Shadow supports Gradle 7.0+ only. Please upgrade.
```

Probably address:
Go to the latest 7.1.2 of the plugin instead of 7.0.0?

Curiosity:
Do you know which pattern of the following is better, or can you explain the difference? idk enough about Gradle
Status quo:
```
buildscript {
  dependencies {
    classpath "gradle.plugin.com.github.jengelman.gradle.plugins:shadow:7.0.0"
  }
}
apply plugin: "com.commercehub.gradle.plugin.avro"
```
another I saw when investigating this:
```
repositories {
    gradlePluginPortal()
}

plugins {
    id "com.github.johnrengelman.shadow" version "7.0.0"
}
```

bbejeck (approved):
LGTM pending the Gradle related tests passing

bbejeck:
@davetroiano the failure is unrelated - the Gradle step passed so feel free to merge

PR #1370, opened 2022-09-15
Navigation UI bug fixed for specific viewport width range

bpurkaystha:
### Description

- For viewport width (992-1100 px), the navigation header was taking only 290px width. (Screenshot-1)
- But, we want it to take full space available for that viewport width range.
- This PR does that with a one-liner CSS change (Screenshot-2)

image

image


scott-confluent (approved):
Thanks for the quick fix!

bbejeck (approved):

PR #1408, opened 2022-09-23
auto-cancel running and queued semaphore pipelines on new push

davetroiano:


davetroiano:
tested this out by manually kicking off build, pushing updated branch, and observing manually kicked off build get canceled

bbejeck (approved):
Nice addition LGTM.

PR #1427, opened 2022-10-19
Add tracking to code copy

hartantothio:
### Description

Send copy event to Segment.

### Staging Docs

http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/track-copy-code

### DIFF
--- a/assets/js/segment.js
+++ b/assets/js/segment.js
@@ -1,4 +1,27 @@
 window.addEventListener('DOMContentLoaded', function () {
+  function copy(node) {
+    let copyText;
+
+    if (navigator.clipboard) {
+      // Modern browser
+      copyText = node.textContent;
+      navigator.clipboard.writeText(copyText);
+    } else {
+      const selection = getSelection();
+      selection.removeAllRanges();
+
+      const range = document.createRange();
+      range.selectNodeContents(node);
+      selection.addRange(range);
+      copyText = range.toString();
+
+      document.execCommand('copy');
+      selection.removeAllRanges();
+    }
+
+    return copyText;
+  }
+
   // If target element is code copy
   function isTargetCodeCopyButton(elem) {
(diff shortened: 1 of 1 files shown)


assets/js/segment.js
```diff
@@ -1,14 +1,40 @@
...
+      const selection = getSelection();
+      selection.removeAllRanges();
+
+      const range = document.createRange();
+      range.selectNodeContents(node);
+      selection.addRange(range);
+      copyText = range.toString();
+
+      document.execCommand('copy');
+      selection.removeAllRanges();
```
rafikalid, line 19: Maybe you need to restore selected ranges from line 11 no?

assets/js/segment.js
```diff
@@ -1,14 +1,40 @@
...
 
   // Determine whether we should track or not
   function shouldTrack(e) {
-    if (e.target.tagName === 'A' || isTargetCodeCopyButton(e.target)) {
+    if (e.target.tagName === 'A') {
       return { target: e.target };
-    } else if (e.target.parentNode.tagName === 'A') {
+    } else if (
+      isTargetCodeCopyButton(e.target) ||
+      e.target.parentNode.tagName === 'A'
```
rafikalid, line 36: Maybe not just parent,  we should check all top parents 
I suggest use:
`if(e.target.closest('a')){ ... }`

hartantothio: I looked at the structure for the site, and it's pretty straightforward. I think using `closest` (it probably has some perf impact) could certainly work, but I'll have to refactor more to accommodate that use case. I'll address that at a separate point.

danshafie (approved):

rafikalid (approved):

bbejeck (approved):
piggybacking on @danshafie and @rafikalid approval

PR #1432, opened 2022-10-26
BACKLOG: Cleaned up dependencies to include clients explicitly

bbejeck:
### Description
The build script only contains a dependency for Kafka Streams.  The application works because KS pulls in the clients as a transitive dependency. Still, since the application does not use Kafka Streams, we should update the dependencies to pull in clients and remove the KS one explicitly.

Internal changes only to the `build.gradle` file no functional changes.


davetroiano (approved):
LGTM

PR #1448, opened 2022-11-08
upgrade gradle-avro-plugin

davetroiano:


bbejeck (approved):
LGTM! Thanks for the fix @davetroiano

PR #1453, opened 2022-11-16
USERBACK: Add emit changes to stream statement

bbejeck:
### Description

Minor internal change the `alerts_raw` stream needs to have `EMIT CHANGES`;


davetroiano (approved):
LGTM!

PR #1455, opened 2022-11-18
[Segment] Scroll depth tracking

TerryDevS:
### Description
[Asana](https://app.asana.com/0/1200268323978783/1203049426090926/f)

### Staging Docs

http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/segment-scrolldepth-tracking

### DIFF
--- a/assets/js/scrollDepth.js
+++ b/assets/js/scrollDepth.js
@@ -2,9 +2,7 @@ const SCROLL_DEPTH_LEVELS = [20, 40, 60, 80, 99];
 
 window.addEventListener('DOMContentLoaded', function () {
   let observer
-  const el = document.getElementsByTagName('body')[0];
-    
-  el.style.position = 'relative';
+  const el = document.body;
 
   const initObserver = () => {
     const cb = (entries) => {
@@ -13,6 +11,7 @@ window.addEventListener('DOMContentLoaded', function () {
     const position = Number(target.dataset.depthMarkerId);
 
       if (isIntersecting) {
+
         analytics.track('Scroll Depth', {
           scrollDepth: position === 99 ? 100 : position // triggering at 99% is more consistent than 100%.
         });
--- a/assets/js/scrollDepth.js
+++ b/assets/js/scrollDepth.js
@@ -1,19 +1,19 @@
 const SCROLL_DEPTH_LEVELS = [20, 40, 60, 80, 99];
 
 window.addEventListener('DOMContentLoaded', function () {
-    let observer
-    const el = document.getElementsByTagName('body')[0];
+  let observer
+  const el = document.getElementsByTagName('body')[0];
     
-    el.style.position = 'relative';
+  el.style.position = 'relative';
 
-    const initObserver = () => {
-      const cb = (entries) => {
-      const [IntersectionObserverEntry] = entries;
-      const { isIntersecting, target } = IntersectionObserverEntry;
-      const position = Number(target.dataset.depthMarkerId);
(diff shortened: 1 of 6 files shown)


assets/js/scrollDepth.js
```diff
@@ -1,19 +1,19 @@
...
+  const initObserver = () => {
+    const cb = (entries) => {
+    const [IntersectionObserverEntry] = entries;
+    const { isIntersecting, target } = IntersectionObserverEntry;
+    const position = Number(target.dataset.depthMarkerId);
 
       if (isIntersecting) {
-       analytics.track('Scroll Depth', {
+        analytics.track('Scroll Depth', {
           scrollDepth: position === 99 ? 100 : position // triggering at 99% is more consistent than 100%.
```
davetroiano, line 17: can position be floating pt? and should this instead be `position > 99 ? 100 : position`

davetroiano (approved):
seems good to me though @hartantothio or @scott-confluent could you please have a look?

assets/js/scrollDepth.js
```diff
@@ -0,0 +1,46 @@
+const SCROLL_DEPTH_LEVELS = [20, 40, 60, 80, 99];
+
+window.addEventListener('DOMContentLoaded', function () {
+  let observer
+  const el = document.getElementsByTagName('body')[0];
```
rafikalid, line 5: why not using just `document.body` ?

assets/js/scrollDepth.js
```diff
@@ -0,0 +1,46 @@
...
+    
+  el.style.position = 'relative';
```
rafikalid, line 7: Maybe doing this in css instead?

assets/js/scrollDepth.js
```diff
@@ -0,0 +1,46 @@
...
+
+  const initObserver = () => {
+    const cb = (entries) => {
+    const [IntersectionObserverEntry] = entries;
```
rafikalid, line 11: I don't see `entries` anywhere else, I prefer doing destruction directly in arguments
`( [IntersectionObserverEntry] ) =>` instead

assets/js/scrollDepth.js
```diff
@@ -0,0 +1,46 @@
...
+    const { isIntersecting, target } = IntersectionObserverEntry;
```
rafikalid, line 12: Do it in args I prefer, it simpler I think
`( [ { isIntersecting, target } ] ) =>`

assets/js/scrollDepth.js
```diff
@@ -0,0 +1,46 @@
...
+    const position = Number(target.dataset.depthMarkerId);
```
rafikalid, line 13: we need this inside the `if` block only, could you please put it there

assets/js/scrollDepth.js
rafikalid: @davetroiano the position here is a hard coded data, not the scroll position it self.

assets/js/scrollDepth.js
TerryDevS: It is simpler but not better for readability IMO. but I can destructure it.

assets/js/scrollDepth.js
TerryDevS: sorry I am not following

assets/js/scrollDepth.js
TerryDevS: sorry I am not following

rafikalid (approved):

PR #1459, opened 2022-11-21
DP-9632: remediate duplicate Semaphore workflows

cchristous:
This repo currently has a Semaphore project that triggers workflows on both pull requests (PR) and all branches. This configuration results in 2 workflow executions when a PR is created or updated. This PR changes the project configuration to use the default behavior for triggers, `[branches, tags, pull_requests]`, and to whitelist branches that match `[master, release]`.

### DIFF
--- a/service.yml
+++ b/service.yml
@@ -8,3 +8,6 @@ github:
 semaphore:
   enable: true
   pipeline_enable: false
+  branches:
+  - master
+  - release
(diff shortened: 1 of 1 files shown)


davetroiano (approved):
just to confirm @cchristous -- if a PR is opened, a single `pr` build will run?  and then if one or more commits gets pushed after that, then a `push` job will get run for each push?

davetroiano:
awesome, thank you for the PR @cchristous

PR #1469, opened 2022-12-05
MINOR: Parallelize the KT build

bbejeck:
### Description

This PR parallelizes the testing blocks.  This will make merging ksqlDB PRs much easier since all three blocks will execute; we won't have to re-run the build and cross our fingers it will make it to the third block.  Also, it should cut down on the build time, which is a good thing.


davetroiano (approved):
thanks! this'll be a timesaver

PR #1487, opened 2023-01-18
upgrade to Java 17

davetroiano:
### Description

[asana](https://app.asana.com/0/0/1203752202076447/f)

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -110,7 +110,7 @@ blocks:
         - name: Tests
           matrix:
             - env_var: JAVA_VERSION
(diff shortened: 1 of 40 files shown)


.semaphore/semaphore.yml
```diff
@@ -168,7 +168,7 @@ blocks:
 
         - name: KStreams Serialization tests
           commands:
-            - sem-version java 11
+            - sem-version java 17
```
bbejeck, line 171: I think we might be able to get rid of this one since it will use the same version of Java for all the tests.

bbejeck (approved):
LGTM assuming build passing

PR #1488, opened 2023-01-18
[TEST DO NOT MERGE] Change only for ksql

bbejeck:
### Description

Change to `ksql` file only ksql block should build


bbejeck:
This build is correct, Semaphore executed only the `ksql` test block - the actual test failure is unrelated

PR #1491, opened 2023-01-18
[TEST DO NOT MERGE] Change for kstream data harness only

bbejeck:
### Description

Changes to a `_data/harnesses/**/kstreams.yml` file should start unit testing and block one builds


bbejeck:
Semaphore executed the correct block; the individual test failure is unrelated

PR #1513, opened 2023-02-22
Update clone script

bbejeck:
### Description

[asana](https://app.asana.com/0/1201951209538300/1203924091714766/f)
This PR updates the clone script to support new types (namely Flink) and is updated to python 3


bbejeck:
changes to `semaphore.yml` and `tutorials.yaml` are formatting-only changes due to being written out by `rtyaml`

tools/clone-tutorial.py
```diff
@@ -0,0 +1,268 @@
...
+new_tutorial_name_parts = str(args.new_name).split('/')
+new_tutorial_name = new_tutorial_name_parts[0]
+orig_tutorial_type = ''
+new_tutorial_type = ''
+if len(orig_name_parts) > 1:
+    orig_tutorial_type = orig_name_parts[1]
+if len(new_tutorial_name_parts) > 1:
+    new_tutorial_type = new_tutorial_name_parts[1]
+if args.debug:
+    print("Names of tutorial to be cloned " + str(orig_name_parts) + '  ' + orig_tutorial_name
```
davetroiano, line 225: style preference / I think more pythonic to interpolate with f-string:
```
print(f"Names of tutorial to be cloned {orig_name_parts} {orig_tutorial_name} {orig_tutorial_type}")
```
there are a few other spots where I think this'd look better than `+`'ing strings or using `.format()`

tools/clone-tutorial.py
```diff
@@ -0,0 +1,268 @@
...
+                         'session-windows/ksql')
+parser.add_argument('--new-name', required=True,
+                    help='The name of the new tutorial if a subtype i.e flink-sql provide path name '
+                         'session-windows/flinksql')
+parser.add_argument('--orig-main-class', help='Name of original main class (applicable for Java/Flink)')
+parser.add_argument('--new-main-class', help='Name of new main class (applicable for Java/Flink)')
+parser.add_argument('--semaphore-test-name', required=True, help='The name of the test for semaphore.yml file')
+parser.add_argument('--permalink', help='The permalink for the tutorial - more descriptive link')
+# parser.add_argument('--json-file', help="The path to a json file containing these configurations")
+parser.add_argument('--debug', default=False, help='The permalink for the tutorial')
```
davetroiano, line 40: optional: use `logging` module rather than `print` and set the level via argument (e.g. see [this](https://stackoverflow.com/a/20663028/5587460) SO)

the primary benefit would then be that rather than having to check and `print` like:
```
if args.debug:
    print("...")
```
it'd be:
```
logging.debug("...")
```
Not a huge benefit today but would be nice to have for maintenance and evolving this

tools/clone-tutorial.py
```diff
@@ -0,0 +1,268 @@
...
+
+print("Updating the tutorials.yml file")
+tutorials_yaml = get_yaml_file(tutorials_yaml_path)
+tutorials_yaml = update_tutorials_yaml(tutorials_yaml, new_tutorial_name, types_to_enable)
+write_yaml_to_file(tutorials_yaml_path, tutorials_yaml)
+print("Now updating front matter entries")
+if args.permalink is not None:
+    update_front_matter(orig_tutorial_name, new_tutorial_name, orig_tutorial_type, new_tutorial_type, args.permalink)
+else:
+    update_front_matter(orig_tutorial_name, new_tutorial_name, orig_tutorial_type, new_tutorial_type)
```
davetroiano, line 268: collapse to one line since permalink arg is optional with default `None`?  i.e.
```
update_front_matter(orig_tutorial_name, new_tutorial_name, orig_tutorial_type, new_tutorial_type, args.permalink)
```

davetroiano (approved):
nice job, big improvement over the bash scripts. some minor points of feedback here. and then we'll get rid of bash scripts in a follow-on, plus update the README at that point, correct?

tools/clone-tutorial.py
```diff
@@ -0,0 +1,268 @@
...
+    if orig_tutorial_type == '' or orig_tutorial_type not in tutorial_types:
+        print("Tutorial type %s is either unknown or must be provided" % orig_tutorial_type)
+        exit(1)
+    elif new_tutorial_type == '' or new_tutorial_type not in tutorial_types:
+        print("Tutorial type %s is either unknown or must be provided" % new_tutorial_type)
+        exit(1)
+
+
+semaphore_path = '../.semaphore/semaphore.yml'
+tutorials_yaml_path = '../_data/tutorials.yaml'
```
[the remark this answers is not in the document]
bbejeck: yeah - I don't why I went with relative paths - everything based on `kt_home` and will work outside of the `tools` dir or inside of it

tools/clone-tutorial.py
bbejeck: yep went with logging - much better than print statements

bbejeck:
@davetroiano all comments addressed

bbejeck:
failure unrelated

PR #1540, opened 2023-04-13
Update AK versions to 3.4.0

bbejeck:
### Description

Part one of two PRs for this [asana](https://app.asana.com/0/1201906632362444/1203406237816996/f) ticket
This PR expanded to cover a little more ground than anticipated.  The reasons for the expanded scope are:

1.  The `3.4.0` version of Kafka Streams deprecated the `CACHE_MAX_BYTES_BUFFERING_CONFIG` in favor of `STATESTORE_CACHE_MAX_BYTES_CONFIG`.  So those configurations needed to be updated
2.  Several Kafka Streams tutorials failed because the application uses a new method available in Kafka-Clients `3.4.0`, but there's a transitive clients dependency pulled in by tutorials using Schema Registry serdes.  Instead of relying on implicit dependencies, I decided that any Kafka Streams tutorial should explicitly declare the required version of Kafka-Clients needed in the `build.gradle` file.   We can update AK dependencies independent of the CP releases by making this change.

### DIFF
--- a/_includes/tutorials/aggregating-average/kstreams/code/build.gradle
+++ b/_includes/tutorials/aggregating-average/kstreams/code/build.gradle
@@ -31,11 +31,11 @@ apply plugin: "com.github.johnrengelman.shadow"
 dependencies {
   implementation "org.apache.avro:avro:1.11.1"
   implementation "org.slf4j:slf4j-simple:2.0.7"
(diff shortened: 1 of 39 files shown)


davetroiano (approved):
LGTM!

PR #1599, opened 2023-07-14
add a second new line so cat func doesn't mess up config

Cerchie:
### Description

no issue, but needed to add a newline to this piece of config, otherwise the cat command in the tutorial messes up the config

### Staging Docs

no docs stages


bbejeck (approved):

bbejeck:
Thanks for the PR @Cerchie !

PR #1604, opened 2023-07-31
Don't format confluent commands as Bash

brianstrauch:
### Description
When CLI commands are formatted as "bash", certain bash-related keywords are highlighted. For example, take a look at step (5) of https://developer.confluent.io/tutorials/how-to-count-messages-on-a-kafka-topic/confluent.html

### DIFF
--- a/_includes/tutorials/aggregating-count/confluent/markup/dev/ccloud-run-consume.adoc
+++ b/_includes/tutorials/aggregating-count/confluent/markup/dev/ccloud-run-consume.adoc
@@ -1,6 +1,6 @@
 Run the following command to start a Confluent CLI consumer to view consume the events that have been filtered by your application:
 
-```
+```plaintext
 confluent kafka topic consume movie-tickets-sold -b --print-key
 ```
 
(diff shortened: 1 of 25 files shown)


davetroiano (changes requested):
@brianstrauch thank you for noticing this and submitting a PR. The syntax highlighter  (hljs) does language detection and for most (all?) of these it'll pick `sql`, e.g., on [this](http://kafka-tutorials-staging.s3-website-us-west-2.amazonaws.com/fix-formatting/how-to-count-messages-on-a-kafka-topic/confluent.html) staged page `create` and `test` are bold:

image

So we have to be explicit... Changing all of these to `` ```plaintext `` or `` ```text `` or `` ```txt `` will do the right thing (see [here](https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md)).

It would be nice if you could turn off language detection and tell highlightjs to use plaintext if no language is specified but I'm not seeing a way to do that

brianstrauch:
@davetroiano Good catch, I'll opt for "plaintext"!

davetroiano (approved):
LGTM

PR #1637, opened 2023-10-10
remove references to Vault

cchristous:
### Description

This removes the references to Vault. Very little of this logic was actually used. The logic that was used was actually no longer necessary as the Semaphore agents are already set up with the necessary environment variables.

The only part I am uncertain about is the `Makefile`. It looks like it is unused because there are nested Makefiles, but I am not sure how that part works. Though, there is no actual logic required in the top level Makefile anymore with the Vault removal.


bbejeck (approved):
LGTM

